PR #18, opened 2017-01-13
Node editor

### LINKED ISSUE #6: Implement Topomath Node Editor (Author)
- [x] User Interface for choosing between QN and EN
- [x] User Interface for Quantity Node Editor
- [x] User Interface for Equation Node Editor
- [x] Setting Up controller for TopoMath Node Editor in common
- [x] Setting Up Author controller
- [x] Handler Stubs

riteshsamala:
Resolves #6. 

New buttons for "Add Quantity, Add Equation "

Controller and Author Controller set up 

Majority of handler stubs 

Master Merged


sachingrover211:
Thanks for the commit. I have few comments:

- [ You have added files for login-base.css and login-form.css. Where are they being used. If they are being used in login.html (if I remember correctly from dragoon), then I don't see login.html in the commit. I don't even see them being added to index.php in the latest code.]
- [ There are a lot of classes in global.css which are not needed by us, atleast not right now. Like for schema. Please add the classes that are needed by you.]
- [ The commit also has the complete math-parser folder which I had added to git-ignore, may be it was missing in your branch and hence it got committed. Please add the code that you have changed and not everything else in the future.]

The idea of us writing the code for topomath and take the ideas from Dragoon is to follow the coding-style used there as, that has worked well for us in the past. We don't want to use everything blindly, as if we wanted to do that then we could have not started from scratch and just made that we saw fit from Topomath. 

Complete css and js files also add network overload when being fetched from the server. Math-parser is already being used from a github branch, so we should not keep a local copy.

In lieu of all these, I would suggest you to revert the previous commit, as there is a lot of extra code. After that when the commit has the code for node-editor only, we can discuss the code added.

jonwetzel:
@riteshsamala I'm going to make you the assignee and the rest of us reviewers.

riteshsamala:
@jonwetzel ok

sachingrover211:
@riteshreddysamala sorry, now that I read the comment again it sounds really rude. That was not intended. I will remove the dojo files that you talked about. I don't know what I was thinking. :)

jonwetzel:
Code updated.  @dharanisakary @sachingrover211 please review.

sachingrover211:
Would it be possible to see a demo of the sorts, so that we can understand what is done and what is left?

riteshsamala:
yes, load index.php. It loads 2 add node buttons. clicking them gives the
quantity and equation node editors.

what has been done in this update :

1. UI design on both editors ( corresponding html, css )
2. controller and con-author have been set up to support these buttons and
view (constructors, some helper functions). listed most of handler
functions for these two editors.
3. corresponding changes in main.js

Furthermore, I can demo at lab, will be there by 9:15.

www/js/main.js
```diff
@@ -24,9 +36,76 @@ define([
...
+
+		//create new model object
+		var givenModel = "";
+	
+		//create new ui configuration object based on current mode ( and activity)
+		var ui_config = "";
+
+		//For now using empty givenModel and ui_config 
+
+		var controllerObject = new controlAuthor(query.m, givenModel, ui_config);
```
sachingrover211: apparently it was to be confirmed... that's why it was missing...

riteshsamala: I will write down a to do comment there for both model and ui parameters. Is there anything apart from that ?

www/js/model.js
```diff
@@ -174,7 +174,7 @@ define([
 			},
 		};
 
-		var obj.authored = lang.mixin({
+		obj["authored"] = lang.mixin({
```
sachingrover211: @riteshsamala shouldnt this be obj.authored as we had discussed. Only the var had to be removed. Was it still causing some issues?

riteshsamala:
yes that is obj.authored,

https://github.com/Dragoon-Lab/topomath/blob/node_editor/www/js/model.js

that is what code shows, where did you see obj["authored"] ?

www/js/main.js
sachingrover211: Its AOK from what I think. I dont know why it was showing me an old commit where it was obj["authored"]. I remembered and rechecked it. Still need to get the hang of this thing.

riteshsamala:
ok if everything looks good, @jonwetzel can you review and merge into
master

--R

PR #36, opened 2017-01-19
Node editor2 

### LINKED ISSUE #23: Node editor UI changes
- [ ] Replace the editable text box with a text area.  
- [ ] Move the undo and check equation buttons to the lower right (to the right of  the "insert above" and the buttons below it).
- [ ] Change kind of quantity box title to: "Optionality"
- [ ] Change: "equation" in Optionality dropdown to "equations"
- [ ] Rename "Show to student" -> "Given to student"

### LINKED ISSUE #31: handle value/ initial value

riteshsamala:
All latest commits of model_ui merged into node_editor 

which also includes handleInitial value (resolves #31), corresponding PM function, type checker module 

also includes prof suggested text changes to quantity editor (resolves #23)


www/index.php
```diff
@@ -153,41 +153,51 @@
...
 		<!-- this is where the menu as well the node editor html code would be kept.
 		Lets follow the hierarchy used earlier.-->
 
-	
 		<!-- Putting Node-Editor -Dialog stuff for demo -->
 		<div class="claro sameedit" data-dojo-type="dijit/Dialog" id="quantityNodeEditor">
 			
 			<div id="studentModelControl" class="fieldgroup">
-				<label style="width:20ex;" for="setStudentNode">Show to student</label>
+				<label style="width:20ex;" for="setStudentNode">given to student</label>
```
jonwetzel: In general all the title text for widgets should start with a capital letter in our dialogs.  Capitalize "G" in "given".

www/index.php
```diff
@@ -153,41 +153,51 @@
...
 				<label for="setName">Variable</label>
 				<input id="setName" data-dojo-type="dijit/form/ComboBox">
-				<label for="selectKind">Kind of quantity:</label>
+				<label for="selectKind">Optionality:</label>
 				<select id="selectKind" data-dojo-type="dijit/form/Select">
 					<option value='defaultSelect'>--Select--</option>
-					<option value='required'>in equation & required</option>
-					<option value='allowed'>in equation & optional</option>
-					<option value='irrelevant'>not in equation</option>
+					<option value='required'>in equations & required</option>
```
jonwetzel: Capitalize "In"

www/index.php
```diff
@@ -153,41 +153,51 @@
...
+					<option value='allowed'>in equations & optional</option>
```
jonwetzel: Capitalize "In"

www/index.php
```diff
@@ -153,41 +153,51 @@
...
+					<option value='irrelevant'>not in equations</option>
```
jonwetzel: Capitalize "Not"

www/index.php
```diff
@@ -261,9 +271,16 @@
 		<div class="claro sameedit" data-dojo-type="dijit/Dialog" id="equationNodeEditor">
 			
 			<div id="studentModelControl2" class="fieldgroup">
-				<label style="width:20ex;" for="setStudentNode2">Show to student</label>
+				<label style="width:20ex;" for="setStudentNode2">given to student</label>
```
jonwetzel: Capitalize "Given"

riteshsamala:
Will check and get back to you

www/js/con-author.js
```diff
@@ -199,6 +223,73 @@ define([
...
+			//Initial value handler for quantity node
+
+			//IniFlag contains the status and initial value
+			var modelType = this.getModelType();
+			console.log("model type is", modelType);
+			var tempIni = dom.byId(this.widgetMap.initial);
+			var tempInival = tempIni.value.trim();
+			console.log("temporary value is", tempInival);
+			var IniFlag = {status: undefined, value: undefined };
+			if(!((modelType === "given") && (tempInival == '') )){
```
sachingrover211: shouldn't we change the given models to authored model as it has been changed in model.js, to keep everything consistent.

www/js/main.js
```diff
@@ -37,75 +34,117 @@ define([
...
+
+            	if((g.y + scrollTop) < topLimit){
+                	//check if bounds inside
+                	if(g.y < topLimit) { // BUG: this g.y should be absolute coordinates instead
+                    	g.y = topLimit;
+                    	node.style.top = topLimit+"px";  // BUG: This needs to correct for scroll
+                	}
+            	}
+            	_model.active.setPosition(id, index, {"x": g.x, "y": g.y+scrollTop});
+			}, true);
```
sachingrover211: I think I skrewed up the indentation here. Would it be possible to open it in sublime and just convert indentation to tabs. 4 indentations to 1 tab space?

www/js/typechecker.js
```diff
@@ -0,0 +1,107 @@
...
+
+		dialogs: [this.myTooltipDialog, this.myTooltipDialog2],
+
+		closePops: function(){
+			array.forEach(this.dialogs, function(dialog){
+				popup.close(dialog);
+			});
+		},
+
+		checkInitialValue: function(nodeID, lastInput){
```
sachingrover211: Since this function is a generally checks whether you have entered number of not. How about we change the function name to isNumber or something? Typechecker file can be used as a general util file, where we can have such utility functions, and have a singleton object. Or we can externalize validations to this file and use the same idea?

www/js/con-author.js
riteshsamala: yes , agreed and changed ti "authored" ( as per model)

www/js/typechecker.js
riteshsamala: This is something we have to discuss . Changing the name is quite simple but I think since this also considers the lastInitialValue , it is not just a number checker we might want to keep it as it is.

www/js/main.js
riteshsamala: Done

riteshsamala:
I tried changing div positions but drawing pane still overlaps a bit into menu bar. I think sachin is the best person to fix this part of UI as he has been working on it @sachingrover211 @jonwetzel . Rest of changes have been made. Please review

www/js/typechecker.js
sachingrover211: I think, it would be better if we modularize this into two separate functions, in case something comes up in the future. One function can be check numeric value and the other can check last input value. Giving it appropriate names and it starts becoming a general library of small utilities or validation functions.

sachingrover211: for singleton idea in JS see http://stackoverflow.com/questions/1479319/simplest-cleanest-way-to-implement-singleton-in-javascript

sachingrover211:
@riteshsamala I found the issue why this happens. The css for these elements is defined by dojo itself. It is happening like this because, every button in the menu has initially display none attached to it. If we remove the display none then it starts showing with correct css, just like in Dragoon. It was working fine in Dragoon only because the help menu and zoom buttons were always visible in Dragoon. I think to fix this we can use visibility css property. It is like display, but it keeps the dom element in place. Just makes it invisible. Or add some menu element which would be always visible like zoom. The code for it can be directly picked from Dragoon.

riteshsamala:
Added visibility property which prevents overlapping of menu bar and following UI

Added two separate functions in typechecker ( checkNumericValue and checkLastInputValue) ... @sachingrover211 Let me know what you think ( I assumed the typechecker as single utility to host both functions and also further similar functions as discussed)

www/js/typechecker.js
```diff
@@ -94,13 +94,39 @@ define([
...
 				return {status: false, errorType: errorType};
-			}else if(inputString == lastInput.value || inputString==""){
+			}
+			else{
+
+				return {status: undefined, errorType: undefined};
+			}
+		},
+
+		checkLastInputValue: function(nodeID, lastInput){
```
jonwetzel: I think this should be renamed "checkLastInitialValue" since it is hard-coded to check the initial value field.

jonwetzel:
Just waiting on @sachingrover211 to review the new version.

sachingrover211:
@riteshsamala I think these changes are great. Thanks

PR #38, opened 2017-01-24
save session and solution

### LINKED ISSUE #10: Implement saving models

dharanisakary:
Changes for saving the model and logging start session. Please let me know for any issues.

Resolves #10 

Accurate diff here: https://github.com/Dragoon-Lab/topomath/compare/master...save_model

### DIFF
--- a/www/login.html
+++ b/www/login.html
@@ -1,0 +1,78 @@
+<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN">
+<html>
+<head>
+	<meta http-equiv="Content-type" content="text/html;charset=UTF-8">
+	<title>Topomath Demo</title>
+
+	<link rel="stylesheet" type="text/css" href="css/login-base.css">
+	<link rel="stylesheet" type="text/css" href="css/login-form.css">
+
+	<script type="text/javascript">
+
+	function validateForm(){
+		var form = document.forms["problem-select"];
+		var x = form.u.value;
+		if (x==null || x==""){
+			alert("Please enter a username.");
+			return false;
+		}      
+		// If group field is empty, remove from form.
+		// Topomath treats the presence of the group field
+		// as a switch for published versus custom problems.
+		if(form.g.value.trim() == ""){
+		  console.log("Deleting group (g) from GET.", form);
+		  document.getElementById("group").removeAttribute('name');
+		}
+		return true;
+	}
+	</script>
+
+</head>
+<body>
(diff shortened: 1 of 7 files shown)


jonwetzel:
Accurate diff here: https://github.com/Dragoon-Lab/topomath/compare/master...save_model

www/login.html
```diff
@@ -0,0 +1,78 @@
...
+					<input id="problem" type="text" name="p" value="rabbits">
+				</div>
+
+				<div class="fieldgroup">
+					<label for="section">Section: </label>
+					<input id="section" type="text" name="s" value="login.html">
+				</div>
+
+				<div class="fieldgroup">
+					<label for="group">Folder:<br> (i.e. group; leave blank for published model) </label>
```
jonwetzel: The fields which used to be called "group" should be called "folder" in topomath.  Please go ahead and change the id of this field to "folder" for consistency.

jonwetzel:
@dharanisakary  @sachingrover211 

I can open rabbits in author mode multiple times and it saves and loads even though I didn't specify a folder.  I'm not sure if it's correct for this to be possible, but it's useful for now since we don't have create node--rabbits is the only thing we can effectively test.

sachingrover211:
I think this is good to go.

dharanisakary:
@jonwetzel , updated the Id of field as 'folder' instead of 'group'

www/login.html
```diff
@@ -0,0 +1,78 @@
...
+	<script type="text/javascript">
+
+	function validateForm(){
+		var form = document.forms["problem-select"];
+		var x = form.u.value;
+		if (x==null || x==""){
+			alert("Please enter a username.");
+			return false;
+		}      
+		// If group field is empty, remove from form.
```
jonwetzel: This comment refers to the old group field as well, and uses "g" as the variable in line 22.  I think that  needs to be "f" instead.  Replace these references to "group" with "folder" (including the comments) and replace "g" with "f" (including in the console log.)

www/login.html
```diff
@@ -0,0 +1,78 @@
...
+					<input id="problem" type="text" name="p" value="rabbits">
+				</div>
+
+				<div class="fieldgroup">
+					<label for="section">Section: </label>
+					<input id="section" type="text" name="s" value="login.html">
+				</div>
+
+				<div class="fieldgroup">
+					<label for="folder">Folder:<br> (i.e. group; leave blank for published model) </label>
```
jonwetzel: You can remove the reference to group; we won't call folders "groups" in any part of topomath.

PR #43, opened 2017-01-25
Node editor2 	handleInitial related apply directives and functions, handle setStudentEqNode, handle setStudentQtyNode

### LINKED ISSUE #31: handle value/ initial value

### LINKED ISSUE #39: handleSetStudentEqNode

### LINKED ISSUE #40: handleSetStudentQtyNode

riteshsamala:
please review

resolves #31 handleInitial related apply directives and functions, 
resolves #39 handle setStudentEqNode, 
resolves #40 handle setStudentQtyNode

### DIFF
--- a/www/index.php
+++ b/www/index.php
@@ -256,12 +256,11 @@
 				<div class="fieldgroup">
 					 <div class="vertical">
 						<div id="equationLabel"></div>
-							<div id="equationInputboxContainer" class="fieldgroup">
-								<label for="equationInputbox">Equation</label>
-									<textarea id="equationInputbox" rows=4 cols=50 data-dojo-type="dijit/form/SimpleTextarea" style="min-height:60px;"></textarea>
-							</div>
+						<div id="equationInputboxContainer" class="fieldgroup">
+							<label for="equationInputbox">Equation</label>
+							<textarea id="equationInputbox" rows=4 cols=50 data-dojo-type="dijit/form/SimpleTextarea" style="min-height:60px;"></textarea>
+						</div>
 					</div>
-
 				</div>
 
 				<div class="fieldgroup" id="algebraic">
--- a/www/index.php
+++ b/www/index.php
@@ -181,7 +181,7 @@
 				<input id="givenToStudentCheckbox" name="markStudentNode" data-dojo-type="dijit/form/CheckBox" checked="false"/>
 			</div>
 			
-			<div id="modelSelectContainer" class="fieldgroup" style="display:none" >
+			<div id="modelSelectorContainer" class="fieldgroup" style="display:none" >
 				<label for="modelSelector">Select Model</label>
 				<select id="modelSelector" data-dojo-type="dijit/form/Select">
 					<option value='correct' selected>Author's Values</option>
--- a/www/index.php
+++ b/www/index.php
@@ -174,7 +174,7 @@
 		Lets follow the hierarchy used earlier.-->
 
 		<!-- Putting Node-Editor -Dialog stuff for demo -->
-		<div class="claro sameedit" data-dojo-type="dijit/Dialog" id="quantityNodeEditor">
+		<div class="claro sameedit" data-dojo-type="dijit/Dialog" id="nodeEditor">
 			
 			<div id="studentModelControl" class="fieldgroup">
 				<label style="width:20ex;" for="setStudentNode">Given to student</label>
@@ -189,9 +189,9 @@
 				</select>
 			</div>
 
-			<div id="nameControl" class="fieldgroup">
-				<label for="setName">Variable</label>
-				<input id="setName" data-dojo-type="dijit/form/ComboBox">
+			<div id="nameControl" class="fieldgroup" style="display: none;">
+				<label for="setVariable">Variable</label>
+				<input id="setVariable" data-dojo-type="dijit/form/ComboBox">
 				<label for="selectKind">Optionality:</label>
 				<select id="selectKind" data-dojo-type="dijit/form/Select">
 					<option value='defaultSelect'>--Select--</option>
@@ -202,16 +202,23 @@
 			</div>
 			
 			<div class="fieldgroup">
-				<div id="descriptionControlAuthor" class="fieldgroup">
+				<div id="descriptionControlAuthor" class="fieldgroup" style="display: none;">
 					<span class="fixedwidth">
 						<div id="authorDescriptionQuestionMark" class="questionMark"></div>
 						<label for="setDescription">Description</label>
 					</span>
 					<input id="setDescription" data-dojo-type="dijit/form/ComboBox">
 				</div>
+				<div id="explanationControlAuthor" class="fieldgroup" style="display: none;">
+					<span class="fixedwidth">
+						<div id="authorExplanationQuestionMark" class="questionMark"></div>
+						<label for="setExplanation">Explanation</label>
+					</span>
+					<input id="setExplanation" data-dojo-type="dijit/form/ComboBox">
+				</div>
 			</div>
 
-			<div class="fieldgroup">
+			<div class="fieldgroup" id="valueUnitsControl" style="display: none;">
 				<!-- adding a div for value field to control its display in UI -->
 				<div id="initialValueDiv" >
 					<span>
@@ -237,68 +244,25 @@
 				</div>
 			</div>
 			
-			<div id="setRootNode" class="fieldgroup" style="">
+			<div id="setRootNode" class="fieldgroup" style="display: none;">
 				<label for ="markRootNode" title ="Mark this node as a root node.">Root:</label>
 				<input id ="markRootNode" name ="markRootNode" data-dojo-type="dijit/form/CheckBox" value="agreed" checked="false"/>
 				<div id="questionMarkRoot" class="questionMark"></div>
 			</div>
 
-			<div id="setDynamicNode" class="fieldgroup" style="">
+			<div id="setDynamicNode" class="fieldgroup" style="display: none;">
 				<label for ="markDynamicNode" title ="Mark this node as a root node.">Dynamic:</label>
 				<input id ="markDynamicNode" name ="markDynamicNode" data-dojo-type="dijit/form/CheckBox" value="agreed" checked="false"/>
 				<div id="questionMarkDynamic" class="questionMark"></div>
 			</div>
 
-			<div class="fieldgroup">
-				<label for="messageBox">Messages</label>
-				<div id="messageBox" class="textscroll" data-dojo-type="dijit/layout/ContentPane"></div>
-
-				<div style="margin-bottom:10px; display:block">
-					<span>&nbsp;</span>
-					<button id="closeButton" type="button" data-dojo-type="dijit/form/Button" iconClass="dijitNoIcon" style="float:right;">Done</button>
-					<button id="deleteButton" type="button" data-dojo-type="dijit/form/Button" iconClass="dijitNoIcon" style="float:left;">Delete Node</button>
-				</div>
-			</div>
-
-		</div><!-- end of quantity node editor -->
-
-		<!-- crisis Alert -->
-		<div class="claro crisisDialog" id="crisisAlertMessage" data-dojo-type="dijit.Dialog" title="Message">
-			<div id = "crisisMessage"> </div>
-			<button id="OkButton" type="button" data-dojo-type="dijit/form/Button">OK</button>
-		</div> 
-		<!-- Equation Node Editor-->
-		<div class="claro sameedit" data-dojo-type="dijit/Dialog" id="equationNodeEditor">
-			
-			<div id="studentModelControl2" class="fieldgroup">
-				<label style="width:20ex;" for="setStudentNode2">Given to student</label>
-				<input id="setStudentNode2" name="markStudentNode2" data-dojo-type="dijit/form/CheckBox" checked="false"/>
-			</div>
-			<div id="selectModelControl2" class="fieldgroup" style="display:none" >
-				<label for="selectModel2">Select Model</label>
-				<select id="selectModel2" data-dojo-type="dijit/form/Select">
-					<option value='correct' selected>Author's Values</option>
-					<option value='given'>Initial Student Values</option>
-				</select>
-			</div>
-
-			<div class="fieldgroup">
-				<div id="descriptionControlAuthor2" class="fieldgroup" style="">
-					<span class="fixedwidth">
-						<div id="authorDescriptionQuestionMark2" class="questionMark"></div>
-						<label for="setDescription2">Explanation</label>
-					</span>
-					<input id="setDescription2" data-dojo-type="dijit/form/ComboBox">
-				</div>
-			</div>
-
-			<div class="ExpressionContainer" id="expressionDiv" style="">
+			<div class="ExpressionContainer" id="expressionDiv" style="display: none;">
 				<div class="fieldgroup">
 					 <div class="vertical">
 						<div id="equationLabel"></div>
 							<div id="nameControl2" class="fieldgroup">
-								<label for="setName2">Equation</label>
-									<textarea id="setName2" rows=4 cols=50 data-dojo-type="dijit/form/SimpleTextarea" style="min-height:60px;"></textarea>
+								<label for="setEquation">Equation</label>
+									<textarea id="setEquation" rows=4 cols=50 data-dojo-type="dijit/form/SimpleTextarea" style="min-height:60px;"></textarea>
 							</div>
 					</div>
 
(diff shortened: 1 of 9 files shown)


riteshsamala:
Hey @sachingrover211 and @jonwetzel can you please review updated code. Please note that I have used initialViewSettings ( which could replace current authorControls () from constructor and uses node type from model) , I made ids to whole blocks and made display = block (except for root and dynamic). Please let me know what u think about block display ( or should I make individual elements inline)

www/index.php
```diff
@@ -237,68 +244,25 @@
 				</div>
 			</div>
 			
-			<div id="setRootNode" class="fieldgroup" style="">
+			<div id="setRootNode" class="fieldgroup" style="display: none;">
```
sachingrover211: There are two kinds of ids used, one which is setRootNode and the other expressionContainer. Wouldn't it be better if we can follow one way of naming these things, and moreover the name of ids dont usually start with verbs as far as I think. It is an element and should be something more of a noun. What do you think? I think same goes for markRootNode.

sachingrover211: In case of buttons it is usually fine because they signify an event and can be named on what would happen if someone clicks them.

www/js/controller.js
```diff
@@ -285,14 +273,147 @@ define([
...
+			// undo color when new value is entered in the equation box widget
+			w.on("keydown",lang.hitch(this,function(evt){
+				if(evt.keyCode != keys.ENTER){
+					w.set('status','');
+				}
+			}));
+		},
+
+		deleteNode: function(id){
+			//Stub to delete node with id by inturn calling drawmodel.deleteNode in main.js
```
sachingrover211: Would it be possible if you add TODO here. This get's highlighted in every code editor, and will ensure that we do not forget anything. There are bunch of comments before as well like need to delete all existing outgoing connections on line 391 which is again a TODO.

www/js/equation.js
```diff
@@ -48,5 +49,132 @@ define([
...
+			var ops = parse.operators();
+			var allowed = {"*": true, "/": true, "variable": true};
+			for(var op in ops){
+				if(ops[op] > 0 && !allowed[op])
+					return false;
+			}
+			return true;
+		},
+		
+        gradient: function(parse, /*boolean*/ monomial, point){
```
sachingrover211: Please fix the indentation here.

www/js/equation.js
```diff
@@ -48,5 +49,132 @@ define([
...
+			// Find the numerical partial derivatives of the expression at
+			// the given point or at a random point, if the point is not supplied.
```
sachingrover211: What is this code for? isSum, isProduct were written to find the labels that we will show in Dragoon. And I dont think we need gradient right now. Is there something I am missing?

www/js/model.js
```diff
@@ -259,14 +276,40 @@ define([
 			},
 			setColor: function(/*string*/ id, /*string*/ color){
 				this.getNode(id).color = color;
-			}
+			},
+			setAuthorStatus: function(/*string*/ id, /*string*/ part, /*string*/ status){
+				// Summary: function to set the status of node editor in author mode.
+				if(!this.getNode(id).authorStatus){
```
sachingrover211: what is this status field? Is it Genus?

www/js/equation.js
riteshsamala: Please check createInputs once, gradient of parse is further used to determine sign ( chooseSign function) which is part of inputs object which is further pushed to node inputs. Let me know if I am missing something here

www/js/model.js
riteshsamala: No it is not genus, it is  a separate field authorStatus for a node ( applied in applyDirectives each time) . Please check applyDirectives in controller.js and updateModelStatus in con-author js

www/index.php
riteshsamala: Do you suggest changing whole set of ids in node editor or few in particular? I think it is a good idea , may be @jonwetzel you have any suggestions in this context for naming ids of node editor elements

www/js/model.js
sachingrover211: Ok I saw the calls. What are the values the authorStatus take, is it for completeness or something?

www/js/equation.js
sachingrover211: We dont need to show signs in Topomath. That is part of Dragoon logic which I think we dont need here. Right @jonwetzel ?

www/index.php
jonwetzel: I like the term "Container" for the divs which just hold other components.  We can id those divs things like "rootNodeToggleContainer", "dynamicNodeToggleContainer"

Buttons can be named after the event which will follow (using a verb), like Sachin suggested.
Checkboxes can be named like "rootNodeCheckbox"
Select boxes and combo boxes can be named like "unitsSelector"
Text inputs/areas can be named "equationInputbox"

www/js/equation.js
jonwetzel: This is correct, we don't have the signs any more, so we can take this function out for now.  @riteshsamala please remove.

jonwetzel:
@sachingrover211 Are you done reviewing the changes Ritesh made?  Let me know, then I'll take a final look.

sachingrover211:
I replied in the meeting that we should remove author status. I cant see the utility of it. If there is a requirement we can then take it.

riteshsamala:
Merged master and added empty messagebox

www/index.php
```diff
@@ -229,77 +230,35 @@
...
+			<div class="equationInputsContainer" id="expressionDiv" style="display: none;">
 				<div class="fieldgroup">
 					 <div class="vertical">
 						<div id="equationLabel"></div>
-							<div id="nameControl2" class="fieldgroup">
-								<label for="setName2">Equation</label>
-									<textarea id="setName2" rows=4 cols=50 data-dojo-type="dijit/form/SimpleTextarea	" style="min-height:60px;"></textarea>
+							<div id="equationInputboxContainer" class="fieldgroup">
+								<label for="equationInputbox">Equation</label>
+									<textarea id="equationInputbox" rows=4 cols=50 data-dojo-type="dijit/form/SimpleTextarea" style="min-height:60px;"></textarea>
```
sachingrover211: I think there is some issue with indentation here? Second thing is the limit of 50 is on size of the equation or the size of the message box?

www/js/con-author.js
```diff
@@ -169,16 +259,69 @@ define([
...
-		handleEquationDescription: function(description){
-			// Summary: Checks to see if the given equation node description exists; if the
-			//		description doesn't exist, it sets the description of the current node.
+			if(!this._model.active.getNodeIDByDescription(description)){
+				this._model.active.setDescription(this.currentID, description);
+				/*
+				logObj = {
+					error: false
+				};
+				*/
```
sachingrover211: This logging is part of the Author PM right? this is not to be written in every handler?

www/js/con-author.js
```diff
@@ -292,36 +447,34 @@ define([
...
 				var studentNodeID = this._model.student.getNodeIDFor(this.currentID);
 				var studNodeInitial = this._model.student.getInitial(studentNodeID);
-				if(modelType == "given"){
-					//if the model type is given , the last initial value is the new student model value
+				if(modelType == "authored"){
+					//if the model type is authored , the last initial value is the new student model value
 					//which in this case is second parameter
 					this._model.active.setInitial(studentNodeID, newInitial);
-					this.updateStatus("initial", this._model.given.getInitial(this.currentID), newInitial);
+					this.updateStatus("initial", this._model.authored.getInitial(this.currentID), newInitial);
```
sachingrover211: There is nothing called initial now. It was updated by value as per the docs. Right @jonwetzel ?

www/js/model.js
```diff
@@ -287,6 +341,12 @@ define([
 			getAuthorID: function(/*string*/ id){
 				return id;
 			},
+			getParent: function(/*string*/ id){
+				return this.getNode(id).parentNode;
+			},
+			getDynamic: function(/*string*/ id){
+				return this.getNode(id).dynamic;
+			},
```
sachingrover211: this was called accumulator in code as per the discussion with Kurt. The functions for this exist by the name of isAccumulator and setAccumulator.

www/js/model.js
```diff
@@ -329,7 +389,7 @@ define([
 			},
 			setExpression: function(/*string*/ id, /*string*/ expression){
 				this.getNode(id).expression = expression;
-			}
+			},
```
sachingrover211: Please remove the ending empty spaces.

www/js/con-author.js
```diff
@@ -148,7 +190,55 @@ define([
...
+			
+			
+			logObj = lang.mixin({
+				type: "solution-enter",
+				nodeID: this.currentID,
+				propoerty: "name",
+				node: name,
+				value: name
+			}, logObj);
+			//this.logging.log('solution-step', logObj);
```
sachingrover211: we need to either find a common place where we can call this logging object or do something else. Otherwise we will keep adding it to every handler and then that would be hard to maintain in the future due to redundancy.

www/js/con-author.js
```diff
@@ -169,16 +259,69 @@ define([
...
+
+			/*
+			this.logging.log('solution-step', {
+				type: "solution-enter",
+				nodeID: this.currentID,
+				property: "kind",
+				node: this._model.given.getName(this.currentID),
+				value: kind
+			});
+			*/
```
sachingrover211: Another logging object

www/js/con-author.js
```diff
@@ -211,12 +354,23 @@ define([
...
-		},
-
-		handleSetStudentEqNode: function(checked){
-			console.log("********************* in handle set student equation node", checked);
+			if(checked){
+				style.set('modelSelectorContainer', 'display', 'block');
+				var studentNode = this._model.student.getNodeIDFor(this.currentID);
+				if(studentNode == null){
+					//to do : to write addStudentNode which has to add a quantity node, pushing this it backlog 
+					//this.addStudentNode(this.currentID);
```
sachingrover211: The to do is supposed to be written like TODO, as this gets highlighted in every code editor be it sublime or vim. For example check the main.js.

www/js/controller.js
```diff
@@ -116,14 +119,149 @@ define([
...
+				// Scroll to bottoms
+				this.domNode.scrollTop = this.domNode.scrollHeight;
+			};
+
+			/*
+			 Add fields to units box, using units in model node
+			 In author mode, this needs to be turned into a text box.
+			 */
+			//to do : selectUnits comes with student mode
+			/*
```
sachingrover211: please change "to do" to "TODO"

www/js/model.js
```diff
@@ -214,7 +242,23 @@ define([
...
 			getInitialNodeString: function(){
 				return obj.initialNodeString;
-			}
+			},
+			getInitial: function(/*string*/ id){
+				var node = this.getNode(id);
+				return node && node.initial;
+			},
+			getOutputs: function(/*string*/ id){
+				// Summary: return an array containing the output ids for a node.
```
sachingrover211: what are the outputs?

www/js/model.js
```diff
@@ -214,7 +242,23 @@ define([
 			},
 			getInitialNodeString: function(){
 				return obj.initialNodeString;
-			}
+			},
+			getInitial: function(/*string*/ id){
+				var node = this.getNode(id);
```
sachingrover211: there is nothing called initial in the data. It is called value as per the design docs

sachingrover211:
I just checked that in Dragoon logging was explicitly done from each handler. @riteshsamala is there any other cleaner way to do logging. Is there a function that is called from each handler. Like in student mode in dragoon this was pedagogical module processAnswer. This was used to do logging as well.

www/index.php
riteshsamala: fixed the indentation, cols here is for the text area for entering the equation

www/js/con-author.js
riteshsamala: For the above three comments as discussed with @sachingrover211 earlier , a separate function which handles the logging for all the author mode handlers is needed. I think we should make a separate card for this for next week. I discussed this with @jonwetzel briefly

www/js/con-author.js
riteshsamala: Oh I see, made changes. Though I don't get this highlighted in sublime, will check it further but made the changes

www/js/con-author.js
riteshsamala: @jonwetzel @sachingrover211  I remember switching between value and initial in dragoon based on node type but Can't remember discussing anything as such in topomath. @jonwetzel let me know if this changes has to be made

www/js/model.js
riteshsamala: discussed with sachin already , used in updateNodes

www/js/model.js
riteshsamala: dharani is making the corresponding changes after discussing with sachin

www/js/model.js
riteshsamala: Done

www/js/con-author.js
jonwetzel: Let's change it to value.

jonwetzel:
This is pending the initial-->value change and for the missing dynamic handler and the review of equation Done Handler.

www/js/controller.js
```diff
@@ -692,6 +700,335 @@ define([
...
+			}
+		},
+
+		substituteVars : function(parse,ignoreUnknownTest){
+		//getDescriptionID is present only in student mode. So in author mode it will give an identity function. This is a work around in case when its in author mode at that time the given model is the actual model. So descriptionID etc are not available. 
+			var mapID = this._model.active.getAuthoredID || function(x){ return x; };
+			var unMapID = this._model.active.getNodeIDFor || function(x){ return x; };
+			var currentNodelist = [];
+			array.forEach(parse.variables(), function(variable){
+					// Test if variable name can be found in given model
```
sachingrover211: This has double indentation

www/js/model.js
```diff
@@ -199,6 +199,11 @@ define([
 				var node = this.getNode(id);
 				return node && node.expression;
 			},
+			getInputs: function(/*string*/ id){
+				// Summary: return an array containing the input ids for a node.
+				var ret = this.getNode(id);
+				return ret && ret.inputs;
```
sachingrover211: inputs were named link in the data object. Links getter and setter are already part of the model

www/js/model.js
```diff
@@ -330,14 +345,19 @@ define([
...
 				});
 			},
 			getDescription: function(/*string*/ id){
 				return this.getNode(id).description;
 			},
+			getAuthoredID: function(id){
+				// Summary: Return any matched given model id for student node.
+				var node = this.getNode(id);
+				return node && node.authoredID;
+			},
```
sachingrover211: Is this equivalent to descriptionID? if that is the case then authoredID won't exist in author model nodes. It would be part of student model nodes only. @riteshsamala is there something that I am missing here?

www/js/controller.js
```diff
@@ -692,6 +700,335 @@ define([
...
+				//input equation in topomath should contain symbol "=" 
+				//both left hand side and right hand side have to be parsed
+				var splitEq = inputEquation.split("=");
+				if(splitEq.length !== 2){
+					directives.push({id: 'message', attribute: 'append', value: 'Incorrect equation syntax'});
+					return null;
+				}
+				var lhsEquation = splitEq[0];
+				var rhsEquation = splitEq[1];
+				console.log(inputEquation, typeof inputEquation, lhsEquation,typeof lhsEquation);
```
sachingrover211: would this be possible to move this part of the code to equation.js. I had created a function called convert in equation.js which takes equation with id's and converts it to equation with variable names. This will make this part reusable in case we need it somewhere else.

www/js/controller.js
```diff
@@ -692,6 +700,335 @@ define([
...
+				var inputNodesList = [];
+				var cancelUpdate = false;
+				var directives = [];
+
+				var widget = registry.byId(this.controlMap.equation);
+				var inputEquation = widget.get("value");
+
+				var lhsInputs = this.substituteVars(parse.lhsParse,ignoreUnknownTest);
+				var rhsInputs = this.substituteVars(parse.rhsParse,ignoreUnknownTest);
+				inputNodesList = lhsInputs.concat(rhsInputs);
```
sachingrover211: there is a function in parser which helps us replace the ids with other strings that we want. Check the function convert in equation.js.

www/js/controller.js
```diff
@@ -692,6 +700,335 @@ define([
...
+							//create node
+							var id = this._model.active.addNode();
+							this.addNode(this._model.active.getNode(id));
+							this.setNodeDescription(id, variable);
+
+							currentNodelist.push({ "id": id, "variable":variable});
+							//get Node ID and substitute in equation
+							var subID2 = unMapID.call(this._model.active, givenID||id);
+							parse.substitute(variable, subID2); //this should handle createInputs and connections to automatic node
+						}/*else{
```
sachingrover211: Yeah substitute is the funciton, but would it be possible to remodularize this so that auto create nodes work and there is a function in equation.js which can be called to replace names with their corresponding ids. It takes an input of the complete equation and replaces them with their ids. So basically you will parse it auto create the new nodes, and then send the equation to that function in equation.js which will return the equation with ids.

jonwetzel:
@sachingrover211 @riteshsamala I started testing this and I'm not all sure how much of it should work at this point, but here's what I noticed:

1. Going directly to index without using login.html throws a javascript error about the message box.
2. Attempting to open a model in a folder causes some unknown javascript error.  Opening rabbits in the null folder opens the model, and appears to be the only way to open a model at this time.
3. Node editor done button appears to do nothing.  X will still close the dialog though.

Let me know if we can merge under these conditions.

riteshsamala:
Regarding done button : it does not work because it should be attached to close editor function which was not part of the previous set of commits.

sachingrover211:
@jonwetzel yeah. The problem not opening issue has been fixed as well. So once this goes to master. I will update my pull requests as well and logging done button as well as the problem opening errors will be fixed.

PR #50, opened 2017-01-31
Create node

### LINKED ISSUE #8: Integrate node creation with draw model

sachingrover211:
Changes for showing a node with when create quantity node is clicked, and the dynamic position code is handled as well.

Resolves #8

### DIFF
--- a/www/js/main.js
+++ b/www/js/main.js
@@ -168,19 +168,28 @@ define([
 			menu.add("createQuantityNodeButton", function(e){
 				event.stop(e);
 				//give a fake id for now
-				var id = "id1";
+				var options = {
+					type: "quantity"
+				};
+				var id = _model.active.addNode(options);
+				console.log("New quantity node created id - ", id);
 				controllerObject.showQuantityNodeEditor(id);	
+				dm.addNode(_model.active.getNode(id));
 			});
 
 			//next step is to add action to add equation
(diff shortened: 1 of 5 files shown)


www/js/main.js
```diff
@@ -170,19 +170,28 @@ define([
 			menu.add("createQuantityNodeButton", function(e){
 				event.stop(e);
 				//give a fake id for now
```
jonwetzel: You can remove this comment now. :)  

Also remove the same comment from line 185, and the dead code from line 190.

www/js/model.js
```diff
@@ -175,7 +218,31 @@ define([
 		};
 
 		obj.authored = lang.mixin({
+			addNode: function(options){
+				obj._updateNextXYPosition();
+				var newNode = lang.mixin({
+					ID: "id" + obj._ID++,
+					genus: "required",
+					accumulator: false,
```
jonwetzel: Why is this field called "accumulator"?  Shouldn't it be "dynamic" instead?

sachingrover211: Professor, in our discussion had said it to be accumulator. This is what is in the data file.

www/js/main.js
sachingrover211: Removed

www/js/model.js
jonwetzel: Ok!

PR #56, opened 2017-02-09
Done Button Changes

dharanisakary:
Hi Jon, 

These are the changes related to Done Button check and alert dialog. Please review and let me know for any issues.

Thanks,
Dharani

### DIFF
--- a/www/js/alert-dialog.js
+++ b/www/js/alert-dialog.js
@@ -48,14 +48,14 @@ define([
 				var b = new Button({
 					label: Object.keys(button)
 				});
-				//console.log("trying aspect",button,typeof button,typeof b.label[0]);
-				aspect.after(button,b.label[0],function(response){
-					console.log("aspect works!");
-				});
+				
+				aspect.after(button,b.label[0],lang.hitch(this, function(response){
+ 					this.close();
+ 				}));
 				handler = on(b, "click", Object.values(button)[0]);
-				this._clickHandlers[button] = handler; 
+				this._clickHandlers[button] = handler;
 				domConstruct.place(b.domNode, d, "first");
-
+				
 			}));
 
 			var dialogContent = "";
@@ -71,6 +71,7 @@ define([
 			registry.byId("alertDialog").hide();
 		},
 		close: function(){
+			console.log("Close Called!! ");
 			if(window.history.length == 1)
 				window.close();
 			else
--- a/www/js/alert-dialog.js
+++ b/www/js/alert-dialog.js
@@ -10,8 +10,9 @@ define([
 	"dojo/query",
 	"dijit/registry",
 	"dojo/html",
+	"dojo/aspect",
 	"dijit/form/Button",
-], function(array, declare, lang, on, dom, domConstruct,  query, registry,html, Button) {
+], function(array, declare, lang, on, dom, domConstruct,  query, registry, html, aspect, Button) {
 	/* Summary:
 	 *				
 	 * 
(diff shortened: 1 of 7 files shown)


www/js/controller.js
```diff
@@ -276,6 +277,78 @@ define([
...
+			
+			if( this._model ){
+				array.forEach(this._model.active.getNodes(), function (node) {
+					if(node.root){
+						hasRootNode = true;
+					}
+					console.log(node);
+
+					// get variables and equations
+					if(node.variable != "" && node.genus && node.genus == "required" && node.type == "quantity"){
```
jonwetzel: Again, please use === and !== instead of == and !=, as a general rule.

www/js/controller.js
```diff
@@ -276,6 +277,78 @@ define([
...
+					obj[_variable] = tmp;
+					return obj;
+				});
+
+				// for each variable if it is not present in any equation, add it to array 
+				var _requiredVariables = [];
+				array.forEach(_usedVariables, function(item){
+					
+					if(!Object.values(item)[0]){	
+						_requiredVariables.push(Object.keys(item)[0]);
```
jonwetzel: A variable is required if the author set its optionality to required.  Otherwise it is not.  We don't seem to be checking its optionality here though, so unless I'm missing something I think there's a bug here where optional nodes will be counted as required even though they're not.

www/js/alert-dialog.js
```diff
@@ -0,0 +1,97 @@
...
+			var cancelLabel = "Cancel";
+			if(cancelTitle && cancelTitle != ""){
+				cancelLabel = cancelTitle;
+			}
+
+			var closeButton = new Button({
+				label: cancelLabel,
+				id: 'cb'
+			});
+			var handler = on(closeButton, "click", lang.hitch(this, function(buttonsArray){
```
dharanisakary: lang.hitch() returns a function that will execute a given function in a given context. So, this works for multiple buttons.

www/js/alert-dialog.js
```diff
@@ -0,0 +1,97 @@
...
+
+			array.forEach(this._buttonsArray.reverse(), lang.hitch(this, function(button){
+				var b = new Button({
+					label: Object.keys(button)
+				});
+				
+				aspect.after(button,b.label[0],lang.hitch(this, function(response){
+ 					this.close();
+ 				}));
+				handler = on(b, "click", Object.values(button)[0]);
```
dharanisakary: This would be [0] because we are using this to get reference to function which needs to be called on click of particular button.

www/js/alert-dialog.js
```diff
@@ -0,0 +1,97 @@
...
+
+		closeDialog: function(){
+			registry.byId("alertDialog").hide();
+		},
+		close: function(){
+			console.log("Close Called!! ");
+			if(window.history.length == 1)
+				window.close();
+			else
+				window.history.back();
```
dharanisakary: There was a requirement "Take as another argument an array of button text and function pairs where a button will be created for each. In addition to calling the function, all buttons should hide the dialog when clicked (so the functions don't have to hide it themselves)."
Mistook the requirement to have close called on every button click after performing its function. Made the necessary changes to call just close dialog after click executes and moved exit topomath code to ExitTopomath stub in main.js

www/js/alert-dialog.js
```diff
@@ -0,0 +1,97 @@
...
+		},
+		end: function(){
+			console.log("End Called");
+			var closeButton = registry.byId('cb');
+			this._clickHandlers[closeButton].remove();
+			array.forEach(this._buttonsArray, lang.hitch(this, function(button){
```
dharanisakary: We can remove eventHandlers in dojo using remove(). https://dojotoolkit.org/reference-guide/1.10/dojo/on.html

www/js/controller.js
dharanisakary: I checked for genus while adding variables. so, only when the the kind is required, that would be checked. The handle for kind is not in place yet in this branch. Therefore, I think this works for optional and irrelevant nodes when added. I can test this once I integrate this code with handlers.

PR #104, opened 2017-03-28
Radio buttons to select variable type

dharanisakary:


www/index.php
```diff
@@ -212,6 +212,18 @@
...
+			<div id="variableTypeContainer" class="fieldgroup" style="display: none;">
+				<input type="radio" data-dojo-type="dijit/form/RadioButton" name="variableType" id="unknownType" value="unknown"/> 
+				<label for="unknownType">Unknown</label> 
+    			
+    			<input type="radio" data-dojo-type="dijit/form/RadioButton" name="variableType" id="parameterType" value="parameter"/> 
+    			<label for="parameterType">Parameter</label> 
+
+    			<input type="radio" data-dojo-type="dijit/form/RadioButton" name="variableType" id="dynamicType" value="dynamic"/> 
+    			<label for="dynamicType">Dynamic</label> 
+
```
sachingrover211: Please fix the indentation here

www/js/model.js
```diff
@@ -268,11 +272,32 @@ define([
 					}, this);
 				}
 			},
+			updateLinks: function(/*string*/ id){
+				var nodes = this.getNodes();
+				var removeId = id + "_initial";
```
sachingrover211: _initial is a string that is defined in model object at the top. Please include it from there. Just see at the top there are getters for it.

sachingrover211: What is the usage of this function? Would it be possible to add some comments to this one. I think this function already existed. From the looks of it, I am not completely sure.

www/js/model.js
```diff
@@ -311,6 +336,15 @@ define([
 				console.log("node added", newNode.ID, newNode.type);
 				return newNode.ID;
 			},
+			updatePositionXY: function(/*string */ id){
+				console.log(obj);
```
sachingrover211: please add some comments for this one as well. I remember you mentioning it why you made this function. But while reviewing I don't remember the specific purpose.

www/js/con-author.js
```diff
@@ -355,20 +354,30 @@ define([
...
-			/*
-			this.logging.log("solution-step", {
-				type: "solution-enter",
-				nodeID: this.currentID,
-				property: "dynamic",
-				node: this._model.given.getName(this.currentID),
-				value: dynamic
-			});
-			*/
+		handleVariableType: function(event){
```
sachingrover211: @dharanisakary as far as the design goes, we have never passed the whole event to these functions. Right? Would it be possible to follow the same idea as for other handlers? Or is there a reason for passing the whole event to a handler, that I am missing?

www/js/con-author.js
```diff
@@ -178,10 +177,10 @@ define([
 			root_check.on('Change', lang.hitch(this, function(checked){
 					return this.disableHandlers || this.handleRoot(checked);
 			}));
-
-			var dynamic_check = registry.byId(this.controlMap.dynamic);
-			dynamic_check.on('Change', lang.hitch(this, function(checked){
-					return this.disableHandlers || this.handleDynamic(checked);
+
```
sachingrover211: Please remove this empty space. Sorry for the finickiness of this review comment.

www/js/con-author.js
```diff
@@ -132,7 +132,7 @@ define([
 		},
 
 		resettableControls: ["variable","description","value","units","equation"],
-		variableNodeControls: ["variable","value","units","kind","dynamic","root"],
+		variableNodeControls: ["variable","value","units","kind","root"],
```
sachingrover211: won't variableType be a part of this?

www/js/model.js
```diff
@@ -438,10 +472,17 @@ define([
...
-				var valueEntered = node.type && node.type == "equation" || node.accumulator || node.value;
+				// variableType and value combined defines node completion
+				// node is complete in following cases
+				// 1. variableType Unknown and no value 
+				// 2. variableType dynamic and value is valid 
+				// 3. variableType parameter and value is valid
+				
+				var valueEntered = node.type && node.type == "equation" || (node.accumulator && node.value) 
+				|| (node.value == "" && node.variableType == "unknown") ||
+				(node.value && node.variableType == "parameter") ;
```
sachingrover211: we are checking node.value twice, would it be possible to change the if condition and say `(node.value && node.variableType != "unknown")`. This makes me think, the dynamic check becomes redundant. @jonwetzel this means I need to change this at many places for UI.

sachingrover211: we are checking node.value twice, would it be possible to change the if condition and say `(node.value && node.variableType != "unknown")`. This makes me think, the dynamic check becomes redundant. @jonwetzel this means I need to change this at many places for UI.

www/js/con-author.js
dharanisakary: No, because, we use this list to initialize all the node fields in initialization part of node editor, and as we do not get the radio buttons by id., instead, we get them by name using dojo query, this field is  handled explicitly in the initialization part of node editor

www/js/con-author.js
dharanisakary: I had used event.target.value in handler to get the value of the checked radio button.

sachingrover211: Ohh.. is it because of the radio buttons?

PR #105, opened 2017-04-05
Logging user events and general infra

### LINKED ISSUE #60: Refactor logging in node editor handlers

sachingrover211:
Resolves #60 and creates a method to handle logging in similar scenario. Only specific errors in the code that we want to handle like (error of wrong format of value only then we need to explicitly call methods logging.js). Otherwise we can use aspect.after and send it to logs from event-logs.js.


www/js/con-author.js
```diff
@@ -248,13 +248,10 @@ define([
 			
 			
 			logObj = lang.mixin({
-				type: "solution-enter",
-				nodeID: this.currentID,
-				propoerty: "name",
-				node: name,
+				propoerty: "variable",
```
dharanisakary: 'property' typo

sachingrover211:
@jonwetzel @dharanisakary review comments incorporated

www/js/controller.js
```diff
@@ -1072,6 +1072,29 @@ define([
...
+		* @params - obj -   object to be logged. Important parameter in the object is property of the node.
+		*                   without which the logged object wont be sent.
+		*           return obj - complete object which should be used by aspect.after to send it to logs.
+		*/
+		logSolutionStep: function(obj){
+			//Stub for logging the messages, updated in con-author and con-student
+			var property = obj.property;
+			// if node ID was not in the object then send node ID as the current ID.
+			// used for auto created nodes.
+			if(!obj.nodeID){
```
jonwetzel: Since we're going to return null if property is null, I think this if statement should get moved into the first clause of the next if statement, and then return obj should get moved to the end of that.  So the if(property) part either returns null, or does all the stuff and returns obj.

www/js/con-author.js
```diff
@@ -944,7 +899,7 @@ define([
 			if(type == "equation"){
 				if( modelType == "given"){
 					registry.byId(this.controlMap.description).set("disabled", true);
-					registry.byId(this.controlMap.description).set("status", '');
+					registry.byId(this.controlMapts.description).set("status", '');
```
jonwetzel: Pretty sure this change is just a typo.

www/js/controller.js
```diff
@@ -1081,19 +1081,18 @@ define([
...
 				obj.node = this._model.authored.getVariable(this.currentID);
 			}
-			if(property){
-				if(!obj.value) obj.value = registry.byId(this.controlMap[property]).get("value");
-				obj.type = this._model.active == this._model.authored ? "solution-enter" : "solution-check";
-			} else {
-				return null;
-			}
+			if(!obj.value) obj.value = registry.byId(this.controlMap[property]).get("value");
+			obj.type = this._model.active == this._model.authored ? "solution-enter" : "solution-check";i
```
jonwetzel: There's an extra "i" at the end of this line...

When we get our build process going, we can make sure you all can run it to catch typos like this automatically. 😄

jonwetzel:
@sachingrover211 In addition to the typo, there are now merge conflicts to fix after the type selector radio button merge.

sachingrover211:
@dharanisakary please check that I have handled the conflict correctly. There was an issue with the handler of variable type, as I had done some changes for the logging in all the handlers.

sachingrover211:
@jonwetzel - if @dharanisakary is ok with merge conflict handling then I will go ahead merge the pull request.

PR #112, opened 2017-04-12
Show prior in equations and show initial node without value

### LINKED ISSUE #100: Prior parsing: Converting IDs to Names
Whenever there is an ID of the form <ID>_Initial, convert it to prior(<name>)

sachingrover211:
Resolves #100 and double node behavior which was closed earlier.


sachingrover211:
@riteshsamala  I know you have reviewed it but I was requesting dharani to check the code as well and it reset your approval as well.

PR #114, opened 2017-04-12
UI Changes

dharanisakary:


jonwetzel:
@dharanisakary merge conflict detected.

sachingrover211:
@riteshsamala I think it would be better to make the functions isAccumulator and setAccumulator as backward compatible. They can do the exact thing like isAccumulator can internally return this.getVariableType(id) == "dyanmic" and similar functioning for setAccumulator it would get the id and set the variableType to dynamic. Because these changes are bound to happen but moving forward to remove what was used previously would be an extra overhead. As this means go change it everywhere and then test all the functionalities if it was missed somewhere. I know it is easy right now. Moreover what we can do is add console.warning to both the functions as deprecated and use variable type instead. We can remove the functions slowly and steadily in the code for pieces that we work on and test them, and then we can remove them completely from the model in time. What do you think @dharanisakary @jonwetzel

sachingrover211:
@riteshsamala I agree to your other suggestion. If we can keep main clean by the end of the project, that would be an achievement in itself. :)

sachingrover211:
Except for one last thing pointed out by @riteshsamala I think rest look good to me. As soon as that is done I will merge this pull request to master. @dharanisakary

PR #117, opened 2017-04-23
prior implementation changes in controller and equation , minor error…

riteshsamala:
…s in model.js (console.warn instead of console.warning)

### DIFF
--- a/www/js/controller.js
+++ b/www/js/controller.js
@@ -865,7 +865,7 @@ define([
 				console.log("Parser error: ", err);
 				console.log(err.message);
 				console.log(err.Error);
-				this.setAuthorStatus(this.currentID,'equation','incorrect');
+				this._model.authored.setAuthorStatus(this.currentID,'equation','incorrect');
 				//error by definition says equation is unacceptable
 				//this._model.active.setEquation(this.currentID, inputEquation);
 				if(err.message.includes("unexpected variable"))
--- a/www/js/controller.js
+++ b/www/js/controller.js
@@ -865,8 +865,7 @@ define([
 				console.log("Parser error: ", err);
 				console.log(err.message);
 				console.log(err.Error);
-				if(this._mode === "AUTHOR")
-					this._model.active.setAuthorStatus(this.currentID,'equation','incorrect');
+				this.changeControlState("equation","authorStatus","incorrect");
 				//error by definition says equation is unacceptable
 				//this._model.active.setEquation(this.currentID, inputEquation);
 				if(err.message.includes("unexpected variable"))
@@ -1069,12 +1068,10 @@ define([
 				if(parseObject.priorError){ //priorError specifically indicates if there is an error where in a non dynamic node/variable is used inside prior function
 					directives.push({id: 'message', attribute: 'append', value: 'Please make a node dynamic before using it in prior function'});
 					directives.push({id: 'equation', attribute: 'status', value: 'incorrect'});
-					if(this._mode === "AUTHOR")
-						this._model.active.setAuthorStatus(this.currentID,'equation','incorrect');
+					this.changeControlState("equation","authorStatus","incorrect");
 				}
 				else{
-					if(this._mode === "AUTHOR")
-						this._model.active.setAuthorStatus(this.currentID,'equation','correct');
+					this.changeControlState("equation","authorStatus","correct");
 				}
 
 				if(directives.length > 0){
@@ -1164,6 +1161,12 @@ define([
 				domClass.remove(dom.byId(buttonID), "active");
 				toolTip.hide(dom.byId(buttonID));
 			});
+		},
+		/** function: changeControlState
+		* this function udpates a control related data (like status, etc) for a specific control(like equation, description)
+		*/
+		changeControlState: function(control, data, value){
+			//this stub is defined individually mode specifically
 		}
 	});
 });
--- a/www/js/controller.js
+++ b/www/js/controller.js
@@ -865,12 +865,11 @@ define([
 				console.log("Parser error: ", err);
 				console.log(err.message);
 				console.log(err.Error);
+				this.setAuthorStatus(this.currentID,'equation','incorrect');
 				//error by definition says equation is unacceptable
 				//this._model.active.setEquation(this.currentID, inputEquation);
 				if(err.message.includes("unexpected variable"))
 					directives.push({id: 'message', attribute: 'append', value: 'The value entered for the equation is incorrect'});
-				else if(err.message.includes("Please make a node dynamic before")) //This case occurs when an equation used prior(node) but that quantity node is not set to be dynamic
-					directives.push({id: 'message', attribute: 'append', value: 'Please make a node dynamic before using it in prior function'});
 				else
 					directives.push({id: 'message', attribute: 'append', value: 'Incorrect equation syntax.'});
 				directives.push({id: 'equation', attribute: 'status', value: 'incorrect'});
@@ -1059,9 +1058,20 @@ define([
 					var dynamicList = parseObject.dynamicList;
 					array.forEach(dynamicList, lang.hitch(this,function(prior){
 						this._model.authored.setVariableType(prior.id,"dynamic");
+						if(this._model.authored.getPosition(prior.id).length === 1)
+							this._model.authored.updatePositionXY(prior.id);
 						console.log("prior id", prior.id);
 						this.updateNodeView(this._model.active.getNode(prior.id));
 					}));
+				}
+
+				if(parseObject.priorError){ //priorError specifically indicates if there is an error where in a non dynamic node/variable is used inside prior function
+					directives.push({id: 'message', attribute: 'append', value: 'Please make a node dynamic before using it in prior function'});
+					directives.push({id: 'equation', attribute: 'status', value: 'incorrect'});
+					this._model.authored.setAuthorStatus(this.currentID,'equation','incorrect');
+				}
+				else{
+					this._model.authored.setAuthorStatus(this.currentID,'equation','correct');
 				}
 
 				if(directives.length > 0){
--- a/www/js/controller.js
+++ b/www/js/controller.js
@@ -865,7 +865,8 @@ define([
 				console.log("Parser error: ", err);
 				console.log(err.message);
 				console.log(err.Error);
-				this._model.authored.setAuthorStatus(this.currentID,'equation','incorrect');
+				if(this._mode === "AUTHOR")
+					this._model.active.setAuthorStatus(this.currentID,'equation','incorrect');
 				//error by definition says equation is unacceptable
 				//this._model.active.setEquation(this.currentID, inputEquation);
 				if(err.message.includes("unexpected variable"))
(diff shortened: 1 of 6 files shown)


www/js/equation.js
```diff
@@ -39,23 +39,45 @@ define([
...
+											array.forEach(nodeList,function(delNode){
+												subModel.deleteNode(delNode.id);
+												console.log("deleted", delNode.id, "because of error in equation parsing")
+											});
+											throw new Error("Please make a node dynamic before using it in prior function");
+											return;
+										}
+										//if the current occurence of the node is part of prior function
+										//it has to be replaced with node_initial in the model
+										expr.substitutePrior(nodeId+"_initial");
```
sachingrover211: Please use subModel.getInitialNodeIDString(), which gives _initial.

www/js/equation.js
```diff
@@ -39,23 +39,45 @@ define([
...
+										// check if the corresponding node has accumulator set (is dynamic)
+										// If not, throw an error indicating the same
+										if(subModel.getVariableType(nodeId) !== "dynamic"){
+											//Also since there is an error with the current node, already created nodes part of the equation should be deleted
+											//nodeList contains the new nodes added to the model which are part of the current equation
+											array.forEach(nodeList,function(delNode){
+												subModel.deleteNode(delNode.id);
+												console.log("deleted", delNode.id, "because of error in equation parsing")
+											});
+											throw new Error("Please make a node dynamic before using it in prior function");
```
sachingrover211: when you throw an error the code flow will stop. Is this intended?

www/js/equation.js
```diff
@@ -69,6 +91,16 @@ define([
                                 var newId = subModel.addNode(newNodeOptions);
                                 nodeList.push({ "id": newId, "variable":variable});
                                 expr.substitute(variable, newId);
+                                if(currentPriorList.length>0){
+									currentPriorList.some(function(eachPrior){
+										if(eachPrior === variable){
+											// In this case along with new node a corresponding prior node has to be created
+											// We store the node data into dynamicList and send to controller where it further makes UI changes for prior node
+											dynamicList.push({ "id": newId, "variable":variable});
+											expr.substitutePrior(newId+"_initial");
```
sachingrover211: same thing here as well use subModel.getInitialNodeIDString().

www/js/equation.js
```diff
@@ -39,23 +39,45 @@ define([
...
+								currentPriorList.some(function(eachPrior){
+									if(eachPrior === variable){
+										//If the current variable is a part of prior function,
+										// check if the corresponding node has accumulator set (is dynamic)
+										// If not, throw an error indicating the same
+										if(subModel.getVariableType(nodeId) !== "dynamic"){
+											//Also since there is an error with the current node, already created nodes part of the equation should be deleted
+											//nodeList contains the new nodes added to the model which are part of the current equation
+											array.forEach(nodeList,function(delNode){
+												subModel.deleteNode(delNode.id);
```
sachingrover211: didnt understand the use case for deleteNode in equations. What was it?

www/js/equation.js
```diff
@@ -39,23 +39,45 @@ define([
...
 				array.forEach(expressions, function(expr){
 					variableList = variableList.concat(expr.variables());
+					var currentPriorList = expr.priors();
 					array.forEach(expr.variables(), function(variable){
 						//This is the case where node names have to be converted to ids
 						//This situation arises from equationDoneHandler
 						//In such situation convert name to id
 						var nodeId = subModel.getNodeIDByName(variable);
 						if(nodeId){
 							expr.substitute(variable,nodeId);
```
sachingrover211: shouldnt the flow of this be a little different. First we call substitute to use variable with nodeId, then we call it separately again for prior. Shouldn't this be called just once?

www/js/equation.js
```diff
@@ -69,6 +91,16 @@ define([
                                 var newId = subModel.addNode(newNodeOptions);
                                 nodeList.push({ "id": newId, "variable":variable});
```
sachingrover211: there aren't tabs here, it has spaces for some reason. Please fix them to tabs using sublime when you are committing it next time. Thanks

www/js/equation.js
riteshsamala: Say the node repeats , a + b = prior(a)*c ( a and prior(a) both are used in the equation) . This case is possible right ?

www/js/equation.js
riteshsamala: So for example a = prior(b), there is an issue with b that it is not a dynamic node and we throw an error but by that time a is added to model in convert function. So if there is an error with equation we have to delete the node 'a' right ?

www/js/equation.js
riteshsamala: ya, we do not further process the equation (if you are referring to "return;" here). Did I miss anything here , any counter argument ?

www/js/equation.js
riteshsamala: ok thanks

www/js/equation.js
sachingrover211: it is done for each variable right? so a and prior(a) would come again? as far as I understand the variables will have a twice in the equation. once normally and once for the prior. Is it correct?

www/js/equation.js
sachingrover211: No don't delete it. We never deleted the variables that were auto created. If the user wanted he would delete it on his own. May be he wants it.

www/js/equation.js
sachingrover211: ok may be it continues in the stack and are we handling this error somewhere, through try catch or something from where it will be called?

www/js/equation.js
riteshsamala: Ohh ok, will make those changes then, this info has to be communicated back to controller and UI node has to be created for those nodes. (Also showing the error message)

www/js/equation.js
riteshsamala: Yeah we are handling it in controller , https://github.com/Dragoon-Lab/topomath/blob/prior_implementation/www/js/controller.js#L872

www/js/equation.js
riteshsamala: I see what you mean by flow, Yeah we can make a check in advance and avoid that situation , will make those changes

www/js/controller.js
```diff
@@ -1050,6 +1053,15 @@ define([
 						this.addNode(this._model.active.getNode(newNode.id));
 						this.setNodeDescription(newNode.id,newNode.variable);
 					}, this);
+					//dynamicList contains those nodes for which prior node UI changes have to be made
+					//Accordingly, make the node dynamic by changing the variable type and setting the accumulator
+					//Also updateNodeView makes sure changes are reflected instantly on the UI
+					var dynamicList = parseObject.dynamicList;
+					array.forEach(dynamicList, lang.hitch(this,function(prior){
+						this._model.authored.setVariableType(prior.id,"dynamic");
+						console.log("prior id", prior.id);
```
dharanisakary: prior node is overlapping with existing nodes. So, instead of just setting variableType, may be this should trigger handleVariableType explicitly

riteshsamala: I think triggering handleVariableType ( triggering event corresponding to changing radio buttons for var type) is not relevant here because dynamicList is collection of multiple prior nodes and all the list nodes have to be updated, let me know what you think

www/js/controller.js
```diff
@@ -865,7 +865,7 @@ define([
 				console.log("Parser error: ", err);
 				console.log(err.message);
 				console.log(err.Error);
-				this.setAuthorStatus(this.currentID,'equation','incorrect');
+				this._model.authored.setAuthorStatus(this.currentID,'equation','incorrect');
```
sachingrover211: this should be active whenever it is used in controller. Because this code will use authored status which we do not want. If possible make this change whereever needed that way it would be easier for us when we work on student mode.

sachingrover211: If it is too much right now we can actually make a card for this right now. I think we will anyway have to move some part of the code from con-author to controller.js when we make con-student, as that is the idea. Only the difference of the two will sit in specific files.

www/js/equation.js
sachingrover211: cool

www/js/controller.js
riteshsamala: There are 3 instances of this and we can use this._mode == AUTHOR check and use ".active" ? we can still make a card as a check like you said when we further made con-student and stabilize the whole controller

sachingrover211: I am good either way. Let's just send this build. I will remember this because when we make student handlers this would be part of the flow. Put same code in controller.js

www/js/equation.js
sachingrover211: what was the final change did you do for this?

www/js/equation.js
sachingrover211: and for this? I remember me and you discussing the two points. But please put the final changes for these two points for future reference.

www/js/equation.js
riteshsamala: as we discussed yesterday, why the current flow is necessary as it handles cases where both a and prior(a) are present

www/js/controller.js
riteshsamala: ok

www/js/equation.js
riteshsamala: For this we are adding an additional priorError to the current object being returned and it is getting processed in createExpressionNodes where it updates directives and sets equation status to incorrect

www/js/equation.js
riteshsamala: Finally, we decided the current flow is correct because
1. if a and prior(a) are in equation expr.substitute replaces a with id 
2. further substitutePrior will replace prior(a) with id_initial
both cases are handled in the flow

www/js/controller.js
```diff
@@ -865,7 +865,8 @@ define([
 				console.log("Parser error: ", err);
 				console.log(err.message);
 				console.log(err.Error);
-				this._model.authored.setAuthorStatus(this.currentID,'equation','incorrect');
+				if(this._mode === "AUTHOR")
+					this._model.active.setAuthorStatus(this.currentID,'equation','incorrect');
```
sachingrover211: this check is already done in model.js. model.active gets authored with the mode is author else it gets student. You dont need to do this explicitly.

www/js/controller.js
```diff
@@ -1068,10 +1069,12 @@ define([
 				if(parseObject.priorError){ //priorError specifically indicates if there is an error where in a non dynamic node/variable is used inside prior function
 					directives.push({id: 'message', attribute: 'append', value: 'Please make a node dynamic before using it in prior function'});
 					directives.push({id: 'equation', attribute: 'status', value: 'incorrect'});
-					this._model.authored.setAuthorStatus(this.currentID,'equation','incorrect');
+					if(this._mode === "AUTHOR")
```
sachingrover211: same thing

www/js/controller.js
```diff
@@ -1068,10 +1069,12 @@ define([
...
+						this._model.active.setAuthorStatus(this.currentID,'equation','incorrect');
 				}
 				else{
-					this._model.authored.setAuthorStatus(this.currentID,'equation','correct');
+					if(this._mode === "AUTHOR")
```
sachingrover211: same thing

sachingrover211:
Discussed with @riteshsamala and @dharanisakary the pull request is good to go.

PR #122, opened 2017-04-28
changes to fix radio buttons not working properly

### LINKED ISSUE #119: Parameter button does cause editor to ask for value
After filling out all the fields of a new quantity node, I click th eparameter button, and nothing happens.  It shoudl open fields for value and units.  when I close the node and reopen it, the menu has reset to be unknown.

sachingrover211:
issue exists only on windows+chrome on certain machines. resolves #119

### DIFF
--- a/www/js/con-author.js
+++ b/www/js/con-author.js
@@ -24,11 +24,12 @@ define([
 	'dojo/dom',
 	'dojo/dom-class',
 	'dijit/registry',
+	'dojo/NodeList-dom',
 	'./controller',
 	"./equation",
 	"./typechecker",
 	"dojo/domReady!",
-], function(array, declare, lang, style, keys, ready, on, memory, aspect, dom, domClass, registry, controller, equation, typechecker){
+], function(array, declare, lang, style, keys, ready, on, memory, aspect, dom, domClass, registry, domList, controller, equation, typechecker){
 
 	// Summary:
 	//			MVC for the node editor, for authors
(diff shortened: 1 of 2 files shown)


www/js/con-author.js
```diff
@@ -24,11 +24,12 @@ define([
 	'dojo/dom',
 	'dojo/dom-class',
 	'dijit/registry',
+	'dojo/NodeList-dom',
 	'./controller',
 	"./equation",
 	"./typechecker",
 	"dojo/domReady!",
-], function(array, declare, lang, style, keys, ready, on, memory, aspect, dom, domClass, registry, controller, equation, typechecker){
+], function(array, declare, lang, style, keys, ready, on, memory, aspect, dom, domClass, registry, domList, controller, equation, typechecker){
```
riteshsamala: What is this module used for exactly ?

www/js/con-author.js
```diff
@@ -178,10 +179,13 @@ define([
 				return this.disableHandlers || this.handleRoot(checked);
 			}));
 
-			var variableTypeToggle = dojo.query("input[type=radio][name=variableType]");
-			variableTypeToggle.on('change', lang.hitch(this, function(){
-				return this.disableHandlers || this.handleVariableType(event);
-			}));
+			debugger;
```
riteshsamala: I think this debugger keyword can be removed now

www/js/con-author.js
sachingrover211: domList is the return object of dojo/query. It is used for defining forEach that I use.

www/js/con-author.js
sachingrover211: done.

sachingrover211:
@riteshsamala should I merge the pull request now?

PR #129, opened 2017-05-05
Refactoring

dharanisakary:


www/js/controller.js
```diff
@@ -1057,11 +1057,7 @@ define([
 					//Also updateNodeView makes sure changes are reflected instantly on the UI
 					var dynamicList = parseObject.dynamicList;
 					array.forEach(dynamicList, lang.hitch(this,function(prior){
-						this._model.authored.setVariableType(prior.id,"dynamic");
-						if(this._model.authored.getPosition(prior.id).length === 1)
-							this._model.authored.updatePositionXY(prior.id);
-						console.log("prior id", prior.id);
-						this.updateNodeView(this._model.active.getNode(prior.id));
+						this.variableTypeControls(prior.id, "dynamic");
```
jonwetzel: @dharanisakary I think making this new function was good (eliminates redundant code), but if the function is going to be used in controller (the super class) then it should be defined here instead of in the subclass (con author).  Does that make sense?

Issue #125, opened 2017-05-02
Rearrange the colors so the more distinct ones get generated first


jonwetzel:
Choose a pallette of very distinct colors and make sure they are the first 10-15.

Issue #121, opened 2017-04-27
To pair quantity nodes with descriptions, use interior color instead of border color


vanlehn:
I've only got a few nodes, and already I can't tell the colors apart on the borders.  Let's use fewer colors, but also use dotted and dashed borders beside the solid ones.  This means that we can't used dotted borders for indicating incomplete nodes.  However, we now have the ability to use icons on student nodes in order to indicate their correctness.  Let's also use an icon (I suggest a ? inside a circle) to indicate that a node is incomplete.

sachingrover211:
@jonwetzel we need to decide how to make the changes. One idea that I had was change the colors to particularly distinct 10 to 15 colors rather than 40 and then try it for sometime.

vanlehn:
Instead of using the color of the interior of a node to pair up only equations with their explanations, lets also color the interiors of quantity nodes to pair them with their descriptions.  The borders of all nodes can be black, and a dotted border can still mean that the node is incomplete.  Second, we should change the shape of the quantity descriptions, which are now rectangles, so that they have rounded ends and thus have the same shape as the quantity nodes.  That will make it easier for students to remember how the pairings work.

jonwetzel:
https://github.com/Dragoon-Lab/topomath/pull/130

PR #130, opened 2017-05-10
changes to fix the node background color and node description

sachingrover211:
Fix for issues - 
* convert quantity border color to background color #121 
* restructure the colors so that nodes are distinctly understandable. #125


sachingrover211:
@jonwetzel @riteshsamala the review for this is still pending.

PR #138, opened 2017-05-18
fixes Remove variable drop down in the editor in author mode #71

riteshsamala:


www/js/con-author.js
```diff
@@ -745,9 +745,6 @@ define([
 				m = new memory({data: units});
 				unitsWidget.set("store", m);
```
jonwetzel:
```
var m = new memory({data: descriptions});
descriptionWidget.set("store", m);
m = new memory({data: units});
unitsWidget.set("store", m);
```
It looks like m isn't ever being re-used.  Could we just re-write this to:

```
descriptionWidget.set("store", new memory({data: descriptions}));
unitsWidget.set("store", new memory({data: units}));
```
?

riteshsamala: Yes true, like discussed left additional comments just in case someone in future might want to explore the capabilities of memory object wrapper

jonwetzel:
@riteshsamala  @dharanisakary  all comments have been addressed, so I will merge this PR.  Thanks!

PR #139, opened 2017-05-18
Adding Build Infrastructure

dharanisakary:


dharanisakary:
I dint face this issue. Can you please try after I add make file?

jonwetzel:
@dharanisakary  I noticed this also creates a node_modules folder in www and it's showing as untracked in GIT.  Please add www/node_modules to the .gitignore

jonwetzel:
@dharanisakary I ran the make file, and some more things installed, but I still get the same error when I do "gulp build".  Could we meet around 2pm today to debug it in my office?

jonwetzel:
We fixed this by upgrading node on my machine to v7.10.0 and np to v4.2.0.

jonwetzel:
@dharanisakary We can require a node and npm version by adding this to package.json:

{ "engines" : { "node" : ">=7.10.0", "npm" : ">=4.2.0" }

If the version that worked for you is earlier than the above, please change your numbers to match. 

See the "engine" section here for documentation: https://docs.npmjs.com/files/package.json

jonwetzel:
@dharanisakary  I forgot, did we get this running on @sachingrover211's machine too?  If so, we can merge if there's no other changes you want to push.

jonwetzel:
Just waiting on @sachingrover211 to confirm this works on his machine too.  

@sachingrover211 I think all you need to do is run `make -f makefile-dependencies` and then test `gulp build` (which does the build) and  `gulp watch` (which should lint your js files each time they are written to, logging the result to console.)

PR #153, opened 2017-06-22
Student Mode - initial changes

dharanisakary:


www/js/con-student.js
```diff
@@ -0,0 +1,352 @@
...
+							style.set('valueInputboxContainer','display','none');
+							style.set('valueInputboxContainer','disabled',false);
+						}
+					}
+				});
+			}			
+			/*
+			 Set color and enable/disable
+			 */
+			array.forEach(this._model.student.getStatusDirectives(nodeid), function (directive) {
```
sachingrover211: is there a reason for this to be here? As far as I remember this was done in apply status directive function.

www/js/controller.js
```diff
@@ -913,137 +957,8 @@ define([
...
-			// fix for bug : https://trello.com/c/bVYAQBKT ~ Sachin Grover
-			/*else if(cancelUpdate){
-				//in case we are not calling the pm then we need to save the equation to the model.
-				//this._model.active.setEquation(this.currentID, inputEquation);
-			}*/
-			//if(true || !cancelUpdate){
-			//return parse;
-			//}
-			//return null; */
-
```
sachingrover211: good job. I really never liked these commented piece of code.

www/js/draw-model.js
```diff
@@ -479,6 +492,47 @@ define([
...
+					nodeClass = " fa-minus";	
+				}else if( attributeCounter !== 0 && attributeCounter === nodeAttributesNumber && nodeAttributesNumber === nodeAttributesNumberExpected /*Add attempt count here*/){
+					nodeClass = " fa-star";
+				}else if( attributeCounter < 0){
+					nodeClass = " fa-times";
+				}else{
+					nodeClass = " fa-check";
+				}
+				return nodeClass;
+
```
sachingrover211: I dont think this code should be here. Draw-model is not supposed to take decision that is the conditions are decided by pedagogical-module and it should tell everyone else constructor or draw-model that show this class. So there can be a map which will tell the class and it will ask draw-model to go do it, same goes for constructor.

www/js/controller.js
```diff
@@ -807,20 +828,42 @@ define([
...
+			// Value is handled when variableType is parameter or dynamic.
+			console.log("********************* in handleVariableType");
+			var _variableType = event.target.value;
+			if(_variableType === this._model.active.getVariableType(this.currentID)){
+				return;
+			}
+			
+			this.variableTypeControls(this.currentID, _variableType);
+			if(this._mode !== "AUTHOR"){
+				this.applyDirectives(this._PM.processAnswer(this.currentID, 'variableType', _variableType));
```
sachingrover211: This was the whole idea of constructor - con-author and con-student architecture. If you have to do this then it would be better to have a function in con-student and an empty function with the same name in constructor. So what would happen it will be called in student and then it wont be called in author mode. and then the idea of setting a field to be disabled is the job of apply directives. Please would it be possible to do these things there only.

www/js/draw-model.js
```diff
@@ -246,6 +255,10 @@ define([
 				innerHTML: innerHTML
 			}, "statemachine-demo");
 
+			if(this._mode !== "AUTHOR"){
+				var nodeStatusClass = this.getStatus(node);
```
sachingrover211: I dont like the mode field which is coming from the request effecting so many things. Let's do one thing, set a field in model and use that? You can call it paranoia but those parameters are supposed to let us know and then we internally handle everything. I mean in future we will have a single point where if we change something and the mode will change and the behaviors as well. I mean it is very unlikely but do you see the point that I am trying to say here? I am open for suggestions on this one.

www/js/main.js
```diff
@@ -53,12 +54,15 @@ define([
...
-			var errorMessage = new messageBox("errorMessageBox", "error", "Missing information, please recheck the query");
+			/*var errorMessage = new messageBox("errorMessageBox", "error", "Missing information, please recheck the query");
 			errorMessage.show();
 			throw Error("please retry, insufficient information");
+			*/
 		}
 	}
 
+	// TO DO : Activity Parameters 
+	// TO DO : UI Parameters
```
sachingrover211: TO DO is to be written as TODO :P

www/js/pedagogical_module.js
```diff
@@ -0,0 +1,217 @@
...
+		obj.push({id: "message", attribute: "append", value: "The value entered for the " + nodePart + " is " + status + "."});
+	}
+
+	function disable(/*object*/ obj, /*string*/ nodePart, /*boolean*/ disable){
+		obj.push({id: nodePart, attribute: "disabled", value: disable});
+	}
+
+	function display(/*object*/ obj, /*string*/ nodeDiv, /*string*/ display){
+		obj.push({id: nodeDiv, attribute: "display", value: display});
+	}
```
sachingrover211: Nicely done!

sachingrover211: But where is the decision making for this table. Like which gets disabled when and which does not get disabled

www/js/pedagogical_module.js
```diff
@@ -0,0 +1,217 @@
...
+				}else{
+					givenID = this.model.student.getAuthoredID(id);
+					console.assert(nodeEditorActionTable[interpretation], "processAnswer() interpretation '" + interpretation + "' not in table ", nodeEditorActionTable);
+					nodeEditorActionTable[interpretation][this.userType](returnObj, nodePart);
+					
+					currentStatus = this.model.authored.getStatus(givenID, nodePart); //get current status set in given model
+					if (currentStatus !== "correct") {
+						for (var i = 0; i < returnObj.length; i++)
+							if (returnObj[i].value === "incorrect") {
+								// handle incorrect result - increment assistance score
```
sachingrover211: I am just going through this commit by commit. I have totally forgotten whether assisstance score is being kept or not. I hope it is.

www/js/pedagogical_module.js
sachingrover211: Because that decision making has to come from this pm module. I hope I will see it in next commits. If so you can ignore the comment.

www/js/pedagogical_module.js
```diff
@@ -197,20 +231,17 @@ define([
 				}else{
 					givenID = this.model.student.getAuthoredID(id);
 					console.assert(nodeEditorActionTable[interpretation], "processAnswer() interpretation '" + interpretation + "' not in table ", nodeEditorActionTable);
-					nodeEditorActionTable[interpretation][this.userType](returnObj, nodePart);
-					
+					nodeEditorActionTable[interpretation][this.userType](returnObj, nodePart, answer);
```
sachingrover211: where is this thing? I mean the nodeEditorActionTable. Why cant I see it where it is defined.

www/js/pedagogical_module.js
```diff
@@ -0,0 +1,336 @@
...
+			feedback: function(obj, part){
+				followUpTasks(obj, part, /*state*/ "incorrect", /*message*/ "incorrect", /*disable*/ false);
+			}
+		},
+		secondFailure:{
+			feedback: function(obj, part){
+				followUpTasks(obj, part, /*state*/ "demo", /*message*/ "secondFailure", /*disable*/ true);
+			}
+		}
+	};
```
sachingrover211: ok here it is. Fine I see it now. Just ignore that earlier comment.

www/js/main.js
```diff
@@ -69,19 +73,63 @@ define([
...
+					//errorMessage.show();
+					throw Error(err);
+				} else {
+					//var errorMessage = new messageBox("errorMessageBox", "error", "This problem could not be loaded. Please contact the problem's author.");
+					//errorMessage.show();
+					throw Error("Model could not be loaded.");
+				}
+			}
+			// This version of code addresses loading errors in cases where problem is empty, incomplete or has no root node in coached mode
+			if (query.m !== "AUTHOR") {
```
sachingrover211: I think using this here is still fine, as main.js is something like our direct connection to that request response. So using the query parameter is fine here. But draw model shouldnt be directly effected by the query parameters. So let's set it in the model.js and use that from now going forward? And with time we can change it at other places as well. I think I have made this same mistake a lot many times.

www/js/draw-model.js
```diff
@@ -479,6 +549,24 @@ define([
...
+				"demo" : "fa-minus",
+				"incorrect" : "fa-times",
+				"correct" : "fa-check",
+				"perfect" : "fa-star"
+			}
+			nodeStatus = this._model.getCorrectness?this._model.getCorrectness(nodeID):"";
+			//Check for perfect node
+			if(this._model.isComplete(nodeID) && this._model.getAssistanceScore(nodeID) === 0 && this._model.getCorrectness(nodeID) === "correct"){
+				nodeStatus = "perfect";
+			}
```
sachingrover211: this is what I mean, these decisions are to be taken by PM and it should tell that nodeStatus is perfect. so show whatever u have for that.

www/js/controller.js
```diff
@@ -640,15 +583,18 @@ define([
...
-					d.getOptions(desc).disabled=exists;
-					if(desc.value == nodeName){
-						d.getOptions(desc).disabled=false;
-					}});
+					
+					if(d.getOptions(desc)) {
+						d.getOptions(desc).disabled=exists;
+						if(desc.value == nodeName){
+							d.getOptions(desc).disabled=false;
+						}
```
sachingrover211: is there a scenario reason here for it to be setting disable for a particular field explicitly? This should come from PM and applyStatusDirective should do it right?

www/js/controller.js
```diff
@@ -807,20 +782,36 @@ define([
...
+			// Value is not allowed when variableType is Unknown
+			// Value is handled when variableType is parameter or dynamic.
+			console.log("********************* in handleVariableType");
+			var _variableType = e.target.value;
+			if(_variableType === this._model.active.getVariableType(this.currentID)){
+				return;
+			}
+			this.variableTypeControls(this.currentID, _variableType);
+			if(this._mode !== "AUTHOR"){
+				this.applyDirectives(this._PM.processAnswer(this.currentID, 'variableType', _variableType));
```
sachingrover211: there should have been a PM for author as well. Which would decide what to say. Even if it was as simple as showing done or not done. But I guess we are too late for that. Anyways I will try to refactor this at some point of time.

www/js/con-student.js
dharanisakary: This is to set attributes of various fields of node on opening node editor
.

www/js/main.js
dharanisakary: Will try to make changes accordingly.

www/js/draw-model.js
dharanisakary: Will try to make changes accordingly.

www/js/controller.js
dharanisakary: This will disable the descriptions of the nodes that are already created in description dropdown

jonwetzel:
Looks good, let's test it on devel.

PR #210, opened 2017-08-19
Student mode

sachingrover211:
Bug fix for given to student #205 
Logging student actions #176

### DIFF
--- a/www/js/con-author.js
+++ b/www/js/con-author.js
@@ -474,6 +474,10 @@ define([
 				return;
 			}
 			this.variableTypeControls(this.currentID, _variableType);
+			this.logSolutionStep({
+                property: "variableType",
+                value: _variableType
+            });
 		},
 
 		handleInputs: function(name){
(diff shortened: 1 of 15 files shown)


www/js/con-author.js
```diff
@@ -814,46 +818,63 @@ define([
...
 					this._model.student.setStatus(newNodeID, "equation" , {"disabled":true,"status":"correct"});
-				}
-				else{
+				}else{
 					this._model.student.setInputs([], newNodeID);
 					this._model.student.setEquation(newNodeID, "");
-					this.errorStatus.push({"id": nodeid, "isExpressionCleared":true});
-					this._model.student.setStatus(newNodeID, "equation" , {"disabled":false,"status":"incorrect"});
+					//this.errorStatus.push({"id": nodeid, "isExpressionCleared":true});
+					//this._model.student.setStatus(newNodeID, "equation" , {"disabled":false,"status":"incorrect"});
```
riteshsamala: why is this set status unnecessary if conversion fails ?? has this been compensated somewhere else ?

www/js/controller.js
```diff
@@ -814,12 +814,8 @@ define([
 				registry.byId(this.controlMap.value).set('value','');
 				this._model.active.setValue(id, '');
 				domStyle.set('valueInputboxContainer','display','none');
-				this.handleValue(null);
+				//this.handleValue(null);
```
riteshsamala: what was this supposed to do actually ? handleValue(null)

sachingrover211: I dont know... it was calling PM with a null value as well. I think @dharanisakary was ensuring the container and value in the model was empty and that happens from the lines above. I didnt want to call handler with the null value explicitly and did testing that it doesnt break anything. I dont think anything wrong is happening with removing this piece of code.

www/js/con-author.js
sachingrover211: the conversion fails here because the equation has nodes that are not part of the model given to the student. So the equation is removed and I dont want it to show red initially so I removed it. Now it shows empty without the red color.

www/js/controller.js
riteshsamala: ok

PR #219, opened 2017-08-29
No feedback

sachingrover211:
No feedback mode changes with a few fixes


sachingrover211:
The above bug has been handled, and correct logging messages were checked. This also initializes the tutor configurations that we will be using going forward.

sachingrover211:
@riteshsamala please let me know the name that you are using for the NOFEEDBACK mode. I will have to make some appropriate changes for that. @jonwetzel please see what the final names should be to follow a proper guideline which will include earlier made student mode as well.

riteshsamala:
@sachingrover211 as discussed yesterday its "NOFEEDBACK"

jonwetzel:
@sachingrover211 Looks like you're going to need to change taskfetcher.php to support multiple student modes.  I suggest you remove the hardcoded reference to STUDENT mode and instead make that clause and if instead of an elseif.

www/js/tutor-configuration.js
```diff
@@ -10,7 +10,7 @@ define(["dojo/_base/lang"], function(lang){
 			]
 		},
 
-		TEST: {
+		NOFEEDBACK: {
 			feedbacMode: "nofeedback",
```
jonwetzel: Typo here: feedbac --> feedback

jonwetzel:
@sachingrover211 Model still doesn't load, though there's no error now.  Looks like there is still a reference to m = EDITOR somewhere:

selection_092

jonwetzel:
One major bug found:

1. Put in an invalid equation.
1. Fix invalid equation.
1. Node now permanently has a red x icon.

jonwetzel:
Bugs with quantities:

1. In no feedback mode, add a quantity node which should be unknown.
1. Select correct description and variable
1. Set type to parameter and enter a value, then click Done on node editor.
1. Open node editor for the same node again.
(Bug 1: variable type has been cleared)
1. Select "unknown" for type
Bug 2: Console error, see screenshot:
selection_095
1. Click done and then re-open node editor a third time. 
Bug 3: "parameter" option is now disabled.

PR #241, opened 2017-09-27
Alphabetically sorting quantities list on canvas

dharanisakary:


www/js/draw-model.js
```diff
@@ -502,9 +503,20 @@ define([
...
+				if(type==="quantity") this.sortQuantityDescriptions();
 			}
 		},
-
+		sortQuantityDescriptions: function(){
+			var quantityParent = dom.byId("quantity-description");
+			var quantityDivs = quantityParent.querySelectorAll(".quantityDescription");
+			var sortedArr = Array.prototype.slice.apply(quantityDivs);
+			sortedArr.sort(function(a,b){
+				return dom.byId(a).querySelector(".descriptionText b").innerHTML.toLowerCase() >  dom.byId(b).querySelector(".descriptionText b").innerHTML.toLowerCase();
```
sachingrover211: Please restructure the code and write the code at appropriate places. The logic for sorting the descriptions should not be written in draw-model. The idea here is that if we write it in model.js the code will be reusable. Else if we want to use this sorted object somewhere, we can not do that as we will need to make draw-model object somewhere else, which defeats the purpose of the whole thing.

jonwetzel:
@dharanisakary merge conflict detected

www/js/model.js
```diff
@@ -981,6 +993,19 @@ define([
...
 				});
+			},
+			getDescriptionsSortedByName: function(){
+				var descriptions = obj.active.getDescriptions();
+				var descNameMap = array.map(descriptions, function (desc) {
+					var authoredID = obj.student.getAuthoredID(desc.value);
+					var _name = obj.authored.getName(authoredID);
+					return {name: _name, description: desc.label, id: desc.value};
+				}, this);
+				descNameMap.sort(function (obj1, obj2) {
```
sachingrover211: These two functions can be merged with a simple check. I don't think there is a need for two separate functions.

Issue #173, opened 2017-07-26
Error handling framework for graph


jonwetzel:
Implements error display for graph

Will be used by pedagogical module too.

PR #242, opened 2017-09-28
Solver feedback

### LINKED ISSUE #177: Feedback on Graph Window for errors and problem completion

sachingrover211:
Changes to show feedback on graph and error messages on graph.
Bug fixes for graph like equations going into infinite loop.

Resolves #177 #173


www/js/user-messages.js
```diff
@@ -0,0 +1,52 @@
...
+	return {
+		get: function(type){
+			var obj = {
+				"graph": graphError,
+				"pm": pedagogical,
+				"app": global
+			}
+
+		 	return obj.hasOwnProperty(type) ? obj[type]: null;
+		},
```
jonwetzel: @sachingrover211 Here's the trailing comma causing the build error.

sachingrover211: I guess I need to learn to see the error message again. So will ask you. I have committed the change. Thanks.

www/js/calculation.js
```diff
@@ -15,29 +15,29 @@ define([
...
-				else
-					system = this.initializeSystem(this._model.authored, id);
-			}
-			var solution;
-			try{
-				system.plotValues = equation.graph(subModel, system, id);
-			} catch(e) {
-				console.error(e);
+
+			if(!system.status.error){
```
dharanisakary: Got a case where system is undefined, and system.status.error was throwing an error in console

dharanisakary:
When I add a unknown quantity node to the model which already has set of equations and try Graph, I get an error in console near initializing Table and Error: <polyline> attribute points: Expected number, "16.50000000 NaN 57.27777778 …" by chart.render().

dharanisakary:
Dropdown should be allowed in the static graph only when there are parameters in the equation. This causes error in the following scenario.
I tried to create a model with 2 equations and 2 unknowns, assigned value to an unknown quantity node in one of the equations. ( Example - a, b are 2 unknown quantity nodes. a=10 is the first equation and b = a + 10 second equation ).
When the model is graphed, in static graph, dropdown is not populated as there are no parameters. Also, as text is allowed in the dropdown, error occurs when text is typed.
(Uncaught TypeError: Cannot read property 'indexOf' of undefined
    at Object.getID (model.js?1508268557798:249)
    at Object.getAuthoredID (model.js?1508268557798:785)
    at Object.renderStaticDialog (render-solution.js?1508268557798:489)
    at Object.<anonymous> (render-solution.js?1508268557798:459)
    at Object.<anonymous> (lang.js?1508268557798:385)
    at Object.dispatcher [as onChange] (aspect.js?1508268557798:101)
    at Object.<anonymous> (_FormWidgetMixin.js?1508268557798:229)
    at lang.js?1508268557798:385
    at Object.<anonymous> (_WidgetBase.js?1508268557798:1200)
    at lang.js?1508268557798:385)

www/js/equation.js
```diff
@@ -361,11 +373,15 @@ define([
 				xvars: [],
 				plotVariables: [],
 				expressions: [],
-				values: {}
+				values: {},
```
jonwetzel: @sachingrover211  Trailing comma is causing a build error.

jonwetzel: @sachingrover211 I'll fix it this time so we can merge this.

PR #252, opened 2017-10-18
Issue236

dharanisakary:


jonwetzel:
@dharanisakary merge conflict detected

www/js/controller.js
```diff
@@ -594,16 +594,27 @@ define([
...
+					var variable_name = this._model.authored.getName(desc.value);
+					var option = {label:variable_name, value: variable_name};
+					if(v.getOptions(option)){
+						var disabled = false;
+						if(exists!== null && model.getName(exists)!== undefined)
+							disabled = true
+						v.getOptions(option).disabled=disabled;
+					}
+					// To reflect the changes to UI
+					v.startup();
```
sachingrover211: add the logic to disable the list in model.js checking which are nodes exist. If not then we should look at why it wasnt working earlier? I dont understand it seems that you are doing the whole logic again. Same logic that has been done just above it.

www/js/controller.js
```diff
@@ -596,24 +596,19 @@ define([
 				var d = registry.byId(this.controlMap.description);
 				var v = registry.byId(this.controlMap.variable);
 				array.forEach(this._model.authored.getDescriptions(), lang.hitch(this, function(desc){
-					var exists =  model.getNodeIDFor(desc.value);
-
-					if( d.getOptions !== undefined && d.getOptions(desc)) {
-						d.getOptions(desc).disabled=exists;
-						if(desc.value == nodeName){
-							d.getOptions(desc).disabled=false;
+					if(this._model.doesStudentNodeExist(desc.value)){
```
sachingrover211: this code will be called for author mode as well. what happens then?

sachingrover211: I am saying as this is in controller.js

dharanisakary: This part is never executed in author mode as model.getNodeIDFor is not defined for author

sachingrover211:
@jonwetzel this PR is really old. Isnt this already happening?

jonwetzel:
The issue was resolved by other code.

PR #259, opened 2017-11-29
User State - Initial Changes

dharanisakary:


www/js/main.js
```diff
@@ -72,6 +73,28 @@ define([
...
+			// adding close callback to update the state for browser message
+			var compatibiltyState = new State(query.u, query.s, "action");
+			errorMessage.addCallback(function(){
+				compatibiltyState.put("browserCompatibility", "ack_" + getVersion());
+			});
+			compatibiltyState.get("browserCompatibility").then(function(res) {
+				if(!(res && res == "ack_" + getVersion())){
+					errorMessage.show();
+				}
+			});
```
sachingrover211: this code has session info, and should be moved to session.js. It was used in Dragoon please check that.

www/js/model.js
```diff
@@ -45,6 +49,17 @@ define([
...
+			setGraphHelpShown: function(_isGraphHelpShown){
+                this.model.properties._isGraphHelpShown = _isGraphHelpShown;
+            },
+            getGraphHelpShown: function(){
+                console.log("model help",this.model.properties);
+                return (this.model.properties._isGraphHelpShown != undefined)?this.model.properties._isGraphHelpShown : false;
+            },
+            setDoneMessageShown : function(_isDoneMessageShown) {
+                this._isDoneMessageShown = _isDoneMessageShown;
+            },
```
sachingrover211: please use proper indentation.  we use tab which is equal to four spaces. If u use spaces then it is not readable at most of the places.

www/js/model.js
```diff
@@ -45,6 +49,17 @@ define([
 					};
 				}
 			},
+			setGraphHelpShown: function(_isGraphHelpShown){
+                this.model.properties._isGraphHelpShown = _isGraphHelpShown;
```
sachingrover211: doesnt state have this property already? what is the need to save it to the model?

www/js/pedagogical-module.js
```diff
@@ -132,6 +133,12 @@ define([
 
 	function message(/*object*/ obj, /*string*/ nodePart, /*string*/ status){
 		// TO DO : Add Hint messages
+		record.increment(status, 1);
+		if(status === "extra" || status === "irrelevant"){
+			status = "incorrect";
+		}else if(status === "lastFailure" || status === "lastFailure2"){
+			status = "incorrect. The correct answer has been given";
+		}
```
sachingrover211: I have already moved this message out to a file. Is there a reason for it to be added again?

www/js/render-solution.js
```diff
@@ -93,6 +102,24 @@ define([
...
+					var content = "<ul>";
+					for(var msg in this._helpText){
+						content += "<li>" + this._helpText[msg] + "</li>";
+					}
+					content += "</ul>";
+					this.helpDialog = new Dialog({
+						"class": 'graphHintUnderLay',
+						"title": "Graph Help",
+						content: content
+					});
```
sachingrover211: can we make the code to less intrusive. What I mean to say is that you have help texts everywhere in all the files. I just saw one in pedagogical_module. now this is in render-solution.js. If possible please move these codes out. U only need to put an event on helpButton which u can do here if you want and then use that to call this piece of code where all the code for messages is kept in one file. Something I tried for logging messages.

www/js/render-solution.js
```diff
@@ -600,12 +627,45 @@ define([
...
+						onBlur: function(){
+							popup.close(problemDoneHint);
+						}
+					});
+					popup.open({
+						popup: problemDoneHint,
+						around: dom.byId('doneButton')
+					});
+					this._model._isDoneMessageShown = true;
+					this.state.put("isDoneMessageShown",true);
```
sachingrover211: please Dharani the idea is not to just write the code. But their has to be a structure to it. This just makes it hard to manage in the future. The same comments here.

www/js/session.js
```diff
@@ -27,7 +27,24 @@ define([
...
+	    if(M[1]=== 'Chrome'){
+	        tem= ua.match(/\bOPR\/(\d+)/)
+	        if(tem!= null) return 'Opera '+tem[1];
+	    }
+	    M= M[2]? [M[1], M[2]]: [navigator.appName, navigator.appVersion, '-?'];
+	    if((tem= ua.match(/version\/(\d+)/i))!= null) M.splice(1, 1, tem[1]);
+	    return {
+			name: M[0],
+			version: M[1]
+		};
```
sachingrover211: u have this code here then u are checking only the browsers in main.js. Then my suggestion would be to put the check here as well and set a variable which will tell you all this and u can keep using that, and can move the check out and just show the message in main.js

www/js/model.js
dharanisakary: This helps us not to make call for state's variables everytime they are needed

www/js/model.js
dharanisakary: Done!

www/js/pedagogical-module.js
dharanisakary: Removed these

www/js/session.js
dharanisakary: Moved the check from main.js to session.js

jonwetzel:
@dharanisakary Merge conflict detected.  Please fix the conflicts and then let reviewers know when revisions are completed.

www/js/render-solution.js
sachingrover211: @dharanisakary please respond to this and other review comments here.

sachingrover211:
@jonwetzel I think you should test this, before merging it. I have done sanity test and the use cases from what I remembered. Since you would have a better idea of the exhaustive cases, please see for any bugs in this. The code is well structured now.

jonwetzel:
@sachingrover211 Can you resolve the merge conflicts which came from the new graph code?  Then I will test this again.

sachingrover211:
@jonwetzel I will do this tonight.

jonwetzel:
@riteshsamala  @sachingrover211  @dharanisakary   I'm getting this error whenever I try to open a new model in author mode on this branch.  It doesn't happen on master.  Unfortunately the trace stops before telling me which file is causing this...

Uncaught Error: declare: base class is not a callable constructor.
    at err (dojo/_base/declare.js?1532561306219:15)
    at declare (dojo/_base/declare.js?1532561306219:794)
    at render-solution.js:20
    at runFactory (dojo.js:1175)
    at execModule (dojo.js:1303)
    at execModule (dojo.js:1294)
    at execModule (dojo.js:1294)
    at execModule (dojo.js:1294)
    at dojo.js:1341
    at guardCheckComplete (dojo.js:1318)

I retested this again on a fresh clone of the repo and same error occurs on this branch.  Any ideas?

jonwetzel:
Closing this PR until we need this feature.

PR #268, opened 2017-12-11
cache buster for state-machine.css

dharanisakary:


jonwetzel:
I heard from Sachin that this wasn't quite finished yet.  I'll wait and test again after it's ready.

PR #304, opened 2018-02-02
Alert Unknown Variables

dharanisakary:


www/js/controller.js
```diff
@@ -927,6 +927,19 @@ define([
 					subModel: this._model.active
 				};
 				returnObj = expression.convert(equationParams);
+				// TODO - Replace this with activity Parameter, if possible
+				if(this._mode === "STUDENT" && returnObj.unknownNodesList && returnObj.unknownNodesList.length > 0){
```
sachingrover211: there is a check in model object for student - this._model.isStudentMode()

sachingrover211: Actually, can we also move this code to equation, that is where node handling and all those things are handled. This portion sends the equation to parse. This should not have any logic for node changing or creating etc. All this is done in convert function and thus should be there. Ohhh, I see why you did this, now lets do this, as far as I see you picked this code from createExpressionNodes, so lets merge this thing. 

Lets put this code in equation.js, then throw an error in this function and then in catch call this piece of code. And please put this copied code into another function and call that function from both the places.

PR #316, opened 2018-02-13
Topomath Build Script

dharanisakary:
Test Path - http://topomath.asu.edu/topomath-devel/login.html


sachingrover211:
@jonwetzel what should be the steps to test this? I think if the build works fine this should be good as, the code is just for including an API right? 

The other question is do you want me to look at something specific in the code as well?

jonwetzel:
@sachingrover211 Just needed a sanity check.  I've read through this and tested it too, looking good.  @dharanisakary  Merging to master.

PR #322, opened 2018-02-16
changes to fix incorrect equation with wrong operator symbols

sachingrover211:


www/js/equation.js
```diff
@@ -135,8 +139,7 @@ define([
 					connections: connections,
 					dynamicList: dynamicList,
 					equation: expressions[0].toString() + " " + this.equalto + " " + expressions[1].toString(),
-					success: !isError,
-					priorError: isError
+					error: isError,
```
jonwetzel: Remove extra ","

www/js/controller.js
```diff
@@ -1000,7 +1000,7 @@ define([
 					}));
 				}
 
-				if(parseObject.priorError){ //priorError specifically indicates if there is an error where in a non dynamic node/variable is used inside prior function
+				if(parseObject.error){ //error specifically indicates if there is an error where in a non dynamic node/variable is used inside prior function
```
jonwetzel: This comment is no longer true, I think.   Now it indicates other types of errors too.  (That's one of the reasons you renamed it I think.)  Please update the comment as needed.

sachingrover211: This comment is true, I updaed it for different reason. Earlier I am using success and priorError for the same variable. So I gave it one generic name. It is still the marked for prior error only.

sachingrover211:
@jonwetzel I have made the appropriate changes. Kindly let me know if there is anything else that I need to do for this. Thanks

jonwetzel:
@sachingrover211 Merge conflict detected.

sachingrover211:
@jonwetzel Merge conflict fixed.

PR #329, opened 2018-03-02
changed display of incomplete nodes

dharanisakary:


sachingrover211:
@jonwetzel  we need to reach a conclusion as to how we want to do this?

www/js/draw-model.js
```diff
@@ -288,6 +288,9 @@ define([
 			}
 
 			if(!this._model.isComplete(node.ID)){
+				if((this._model.isStudentMode() && this._feedbackMode == "nofeedback") || !this._model.isStudentMode()){
```
sachingrover211: First of all,
This if check can be simplified to - !this._model.isStudentMode || this._feedback == "nofeedback"

which I think is what u are doing here - either it is not student mode here or if it is then feedbackmode is nofeedback.
So basically editor is in author mode or it is nofeedback mode.

I think you can achieve this from config as well, keep the configuration same for test and author, and thus this will be way more simpler.

PR #332, opened 2018-03-13
Fix for author mode parameter node looking like unknown when changing Variable name

dharanisakary:


dharanisakary:
Updated the card with replication steps

PR #334, opened 2018-03-20
changes to ensure that model is saved after node is deleted.

sachingrover211:


sachingrover211:
@dharanisakary that is the reason it is not being saved. If we do save mdoel before the node is deleted, then the saved model is not completely updated, and hence there is a delete bug coming.

sachingrover211:
@dharanisakary can you tell me the specific scenario here that you are pointing out? Some steps to reproduce the bug that you mention here?

sachingrover211:
@dharanisakary would it be possible to say the specific scenario here, like some steps to reproduce what you are saying?

dharanisakary:
@sachingrover211 , below are the steps for error scenario - 
1. Create a new model
2. Create a new unknown quantity
3. Mark it as Given to Student
4. Close the node
5. Delete the node. This causes error - 
Uncaught TypeError: Cannot read property 'attemptCount' of null
    at Object.setAttemptCount (model.js?1522711282461:776)
    at Object.removeStudentNode (con-author.js?1522711282461:923)
    at Object.updateAssignedNode (con-author.js?1522711282461:1109)
    at Object.<anonymous> (main.js?1522711282461:319)
    at Object.dispatcher [as deleteNode] (aspect.js?1522711282461:101)
    at Object.<anonymous> (draw-model.js?1522711282461:328)
    at Object.onClick (lang.js?1522711282461:385)
    at Object.onItemClick (_MenuBase.js?1522711282461:318)
    at HTMLTableRowElement.<anonymous> (_MenuBase.js?1522711282461:101)
    at HTMLTableSectionElement.<anonymous> (on.js?1522711282461:263)

PR #335, opened 2018-03-23
Fix position of nodes

dharanisakary:


sachingrover211:
@dharanisakary  how is this to be tested?

I tried using login.html fix position, but none of the options of the checkbox is working. Although one of the cases work in the weird ways. When the Fix position checkbox is **checked**, the nodes go to their  positions as per the author mode, but they are still moveable. When fix position checkbox is **unchecked**, the equation nodes do not move to author positions and quantity nodes move to author positions, and all the nodes are moveable.

I think I am missing the testing scenario, but please also check the behavior with the fix node checkbox in login.html.

dharanisakary:
So, the card was opened to change the behavior to "Assign author position to the nodes in student mode if the option was checked, else position nodes in natural ordering in student mode". Previously, the checking the option locked the nodes to canvas.

PR #346, opened 2018-04-20
Graph restructuring

sachingrover211:
Complete graph restructuring + adding solution when author graph does not have solution for color by numbers.


jonwetzel:
@sachingrover211 In the Graph vs Parameter tab of the graph window, the user can type in the dropdown box used to select the x-axis variable.  I think that's a bug.  Change it from a combo box to a plain dropdown so the user can't type in it anymore.

jonwetzel:
@sachingrover211 Possibly related to the first bug: I have a system with 3 variables in one equation (d = r*t).  r and t are parameters, d is unknown.  In the graph vs parameter tab, I can plot d vs r but not d vs t.   Nothing happens when I click the dropdown button to try to switch from r to some other variable.

jonwetzel:
We'll raise these as a separate issue.  Merging this one.

PR #348, opened 2018-05-04
Topomath side changes for overdetermined system

sachingrover211:
Tutor side changes for incorporating new error messages and solver updates for preprocessing the equations.

### DIFF
--- a/www/js/equation.js
+++ b/www/js/equation.js
@@ -2,7 +2,7 @@ define([
 	"dojo/_base/array",
 	"dojo/_base/lang",
 	"parser/parser",
-	"solver/newton-raphson"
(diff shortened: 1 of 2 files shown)


jonwetzel:
@sachingrover211 The author mode is complete, so I was able to test it out there.  Cases with fewer unknowns seem to work fine, but when I made extra equations, it shows an error message on hitting graph/table.  

The system: 
selection_131

The error displayed: 
selection_129

The console:
selection_132

The solver doesn't appear to be throwing any error, so I wonder if the bug is actually in the tutor instead.  Please try it out and let me know.

jonwetzel:
@sachingrover211 That fixed it.  Thanks!

PR #357, opened 2018-05-21
qty node refactor author mode

riteshsamala:


sachingrover211:
@riteshsamala would it be possibl to attach it to an issue, so that I can know what was the idea behind the changes.

riteshsamala:
@sachingrover211 Connected. Thanks !

PR #369, opened 2018-07-25
Schema table json updated (extra field added), drt changed to rate

riteshsamala:


jonwetzel:
I didn't realize the slot names were already in the mapping field.  Most of this PR is therefore unnecessary.  I'll make the DRT rename change elsewhere.

Issue #360, opened 2018-06-07
In Author mode, validate the new entity field value


jonwetzel:
If the entity field is exited and has invalid characters, print a message and color the entity field red.

jonwetzel:
Resolved by #374

Issue #345, opened 2018-04-06
Author: Schema & Entity fields


vanlehn:
Update equation editor to have these fields: 
https://www.gomockingbird.com/projects/esbufe3/4gXVnC

- [x] Add Schema and Entity fields like in mockup
- [x] Change description field into a one line text box.
- [x] Populate schema dropdown using values from the schema table, in order of json.  Default value is "\<Select a schema\>" like other dialogs.
- [x] Save values author enters into the model when user clicks done or check equation.
- [x] Hint for entity should say "Enter a list of valid entity names, separated by semicolons"
- [x] The hint icon for schema opens the schema table in a new modal instead of showing a tooltip.
- [x] Hide Add quantity button from Author mode
- [x] If the entity field is exited and has no valid names inside, print a message and color the entity field red.  (Moved to https://github.com/Dragoon-Lab/topomath/issues/360)

jonwetzel:
Resolved by #374

Issue #339, opened 2018-04-03
Authoring eqn explanations in <relationship><name> format


vanlehn:
Currently, authors can enter any text as the explanation of an equation.  The change is to have them enter a set of strings of the form <relationship><name>, where <relationship> is any string that lacks a space, and <name> matches (case doesn't matter) a known schema name.   For instance, suppose the problem is "George takes a bus out of town for 15 minutes before he realizes he forgot his laptop and gets off the bus.  He walks back at a speed of 3 mph for 3 hours.  How fast did the bus travel?"   This requires two applications of the DRT schema.  For one, the author might enter
bus DRT; outbound DRT; first-leg DRT

For the other, the author might enter
walking DRT; walk DRT; inbound DRT; return  DRT; second-leg DRT

jonwetzel:
Old version:

Currently, authors can enter any text as the explanation of an equation.  The change is to have them enter a SET of strings of the form entity.name, where entity is any string that lacks a space, and name matches (case doesn't matter) a known schema name.   For instance, suppose the problem is "George takes a bus out of town for 15 minutes before he realizes he forgot his laptop and gets off the bus.  He walks back at a speed of 3 mph for 3 hours.  How fast did the bus travel?"   This requires two applications of the DRT schema.  For one, the author might enter
bus.DRT; outbound.DRT; first-leg.DRT

For the other, the author might enter
walking.DRT; walk.DRT; inbound.DRT; return.DRT; second-leg.DRT

PR #374, opened 2018-08-07
topomath eq editor updates #339, #345, #360

riteshsamala:


www/js/con-author.js
```diff
@@ -795,6 +890,22 @@ define([
...
 						this.applyDirectives(this.authorPM.process(this.currentID, 'equation', this._model.authored.getEquation(this.currentID), true));
 				}
+				//sessionStorage contains the schema table, load options from schema table
+				var schemasList = [];
+				var schemaTabOb = JSON.parse(sessionStorage.getItem("schema_tab_topo"));
+				for(var parSchema in schemaTabOb){
+					schemaTabOb[parSchema].forEach(function(atomSchema){
+								var sname = atomSchema["Name"];
+								var obj = {value: sname, label: sname};
+								schemaWidget.addOption(obj);
```
jonwetzel: There's a bug with this: we add the options every time this function is called, but we never delete them.  So an extra copy of the schemas is being added every time the node editor opens.  

Perhaps adding the options should be done once when the tutor first loads (in index.php or main.js?) instead of when the node editor opens, because we don't expect the list of schemas to change during a session.

www/js/controller.js
```diff
@@ -412,9 +415,9 @@ define([
 			}));
 
 			 //event handler for quantity node description field
```
jonwetzel: Update comment to match code.

www/js/equation.js
```diff
@@ -43,6 +43,8 @@ define([
 			try{
 				array.forEach(eqs, function(eq, count){
 					expressions[count] = Parser.parse(eq);
+					console.log("expr array",expressions);
```
jonwetzel: Remove this console.log if no longer useful.

www/schema-table.json
```diff
@@ -7,7 +7,8 @@
 				"PA": "Part A",
 				"PB": "Part B"
 			},
-			"Comments": "The whole is equal to the sum of two parts."
+			"Comments": "The whole is equal to the sum of two parts.",
+			"slot names": "Whole, Part A, Part B"
```
jonwetzel: As previously discussed, the slot names field is redundant with the Mapping.  Go ahead and remove all "slot names" fields.

riteshsamala:
I have updated the code in latest commit on current branch. It also includes enhancements. I also found one bug while testing, entity text box was accepting empty entity input value before semicolon like ";entity1". Fixed it as well.

Issue #341, opened 2018-04-03
In author mode, auto-create variable names


vanlehn:
When the author has created a new equation node and entered a syntactically acceptable set of explanations, each of the form <relationship> <name>, then 

1.  First, topomath looks up the schemas corresponding to the <names> and insures that the schema's generic equations are isomorphic.   That is, the equation for DRT is D=R*T and the equation for UnitCost is Total=UC*Units, so they are isomorphic.  On the other hand, the equation for PW2 is W=P1+P2 but the equation for headwind is LandSpeed=AirSpeed - Headwind, so they are not isomorphic.  If the equations are not all isomorphic, then the author gets an error message.

2.  Next, TopoMath defines one quantity node per variable in the schemas' equations.  The variables are assigned all the corresponding names.  Thus, if the author has "bus.DRT" and "outbound.DRT" as the explanations for the equation, then Topomath creates 3 quantity nodes, with the following sets of names {"bus.D", "outbound.D"}, {"bus.R", "outbound.R"} and {"bus.T", "outbound.T"}.  

3.  TopoMath displays in the equation control the equation corresponding  to the first explanation.  For example, if "bus.DRT" is first, then it displays "bus.D = bus.R * bus.T"

jonwetzel:
Old version, keeping in a comment just in case.


 When the author has created a new equation node and entered a syntactically acceptable SET of explanations, each of the form entity.name, then 

1.  First, topomath looks up the schemas corresponding to the names and insures that the schema's generic equations are isomorphic.   That is, the equation for DRT is D=R*T and the equation for UnitCost is Total=UC*Units, so they are isomorphic.  On the other hand, the equation for PW2 is W=P1+P2 but the equation for headwind is LandSpeed=AirSpeed - Headwind, so they are not isomorphic.  If the equations are not all isomorphic, then the author gets an error message.

2.  Next, TopoMath defines one quantity node per variable in the schemas' equations.  The variables are assigned all the corresponding names.  Thus, if the author has "bus.DRT" and "outbound.DRT" as the explanations for the equation, then Topomath creates quantity nodes, with the following sets of names {"bus.D", "outbound.D"}, {"bus.R", "outbound.R"} and {"bus.T", "outbound.T"}.  

3.  TopoMath displays in the equation control the equation corresponding  to the first explanation.  For example, if "bus.DRT" is first, then it displays "bus.D = bus.R * bus.T"

jonwetzel:
[...]

PR #380, opened 2018-08-16
equation editor update 2: variables slot container, dynamic comboboxe…

riteshsamala:
…s based on schema and data populated based on schema mappings

### DIFF
--- a/www/js/con-author.js
+++ b/www/js/con-author.js
@@ -25,13 +25,15 @@ define([
 	'dojo/dom-class',
 	"dojo/dom-style",
 	'dijit/registry',
+	'dijit/form/ComboBox',
 	'dojo/NodeList-dom',
+	'dojo/html',
 	'./controller',
 	"./equation",
 	"./typechecker",
 	"./popup-dialog",
 	"dojo/domReady!"
-], function(array, declare, lang, style, keys, ready, on, memory, aspect, dom, domClass, domStyle, registry, domList, controller, equation, typechecker, popupDialog){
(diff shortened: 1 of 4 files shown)


riteshsamala:
This will handle #341

www/js/con-author.js
```diff
@@ -1234,6 +1237,111 @@ define([
...
+					}
+				}
+				sessionStorage.setItem("slot_number_map", JSON.stringify(numberMap));
+			}
+			//now we have slot number map, we need to check the map and return appropriate vars
+			var numGenOb = JSON.parse(sessionStorage.getItem("slot_number_map"));
+			var numList = Object.keys(numGenOb);
+			var slotVars = Object.keys(slotMap);
+			var nextNum;
+			if(numList.length < 1){
```
jonwetzel: The implementation seems overly complex.  Couldn't we just do something like this?

```
var nextSubscript = 1;
var finished = false;
while (!finished) {
	if (nextSubscript in numGenOb){
		// This subscript is in use by some variable, but maybe not one in the current schema
		if (isSubscriptInUse(nextSubscript,slotVars,numGenOb)){
			// This subscript is in use, increment and try again
			nextSubscript++;
			continue;
		} else {
			// The subscript was in the map, but not in use by any of this schema's variables
			finished = true;
			continue;
		}
	} else {
		// The subscript isn't in the map yet
		finished = true;
	}
}
return nextSubscript;
},

/* 
	isSubscriptInUse
	Given a subscript, list of variables, and the numGenOb
	Return true if the subscript is already in use by any of the variables
*/
isSubscriptInUse: function(subscript,slotVars,numGenOb){
	slotVars.some(function(v){
					return numGenOb[subscript].includes(v);
				  }
		}
	}
}

```

Issue #366, opened 2018-07-24
Update the equation box based on the variable slot values


jonwetzel:
The equation box is now disabled like the description box.  It updates automatically based on the variable names entered in the combo boxes.

- [ ] Disable the equation box
- [ ] Hide the equation editing controls (they will come back for the custom equation schema, so do not delete them or their code yet), including check equation.
- [ ] Whenever a variable combo box updates, edit the equation to match the new variable names.  Also do this when the boxes are set after the user selects a schema.

Issue #368, opened 2018-07-25
Author mode: Equation node on-close behavior


jonwetzel:
(The check equation button has been hidden until we add custom equation schema, so closing the node is the only time we need to update the model.)

When an equation node is closed: 
- [ ] If all variable slots have a valid (non-blank) name, then the equation in the model is updated
- [ ] If some variable slot has an invalid name, the equation node should not close.  Instead it should show feedback in the message box and remain open.
- [x] Any variable nodes which do not exist are created
- [x] The updated model is saved to the database (already happens)

PR #383, opened 2018-08-23
Eq editor update3

riteshsamala:
handles #366


jonwetzel:
According to @riteshsamala this bug will get fixed in the PR for issue #368

Issue #379, opened 2018-08-13
On open, the equation editor should update variables section based on the model


jonwetzel:
On opening a new equation node:
- [ ] All fields should be reset and no variables slots should be present.

On opening an existing equation node:

- [ ] If schema is set, dynamically generate the variable slots for that schema.
- [ ] If equation is also set, fill in the variable names with the ones from the equation in the model (note: some may be blank)
- [ ] If schema is not set, then there will be no variable slots visible until the user selects a schema.  Ignore the equation in this case.  (It should be blank anyway.)

PR #384, opened 2018-09-05
#368 equation node on close behavior, #379 updating equation editor b…

riteshsamala:
…ased on model, bug: updating new equation editor on add equation

PR Todos: 
- [x] Make it so the node editor done button doesn't close the node if the equation has a syntax error--instead display the message in the message box and leave the editor open so the author can fix it.

### DIFF
--- a/www/js/con-author.js
+++ b/www/js/con-author.js
@@ -173,7 +173,7 @@ define([
 
 		resettableControls: ["variable","description","value","units","equation"],
 		variableNodeControls: ["variable","value","units","kind","root"],
-		equationNodeControls: ["inputs","equation","schemas"],
+		equationNodeControls: ["equation","schemas","entity"],
 		commonNodeControls: ["setStudent","modelType","description"],
 
 		controlMap: {
@@ -598,7 +598,6 @@ define([
 						message: "parse error"
 					};
 				}
-				console.log("directives are", directives);
 				this.applyDirectives(directives);
 				this.createExpressionNodes(returnObj, true);
 				return directives; 
@@ -727,10 +726,6 @@ define([
 			var equationDescriptions = [];
 			var units = [];	
 			var schemaList = [];
-
-			var desc = this._model.authored.getDescription(nodeid);
-			registry.byId(this.controlMap.description).set('value', desc || '', false);
-
 			// Initialize student node checkbox
 			//this is common to both quantity and equation nodes in topomath
 				var givenNode = this._model.authored.getNode(nodeid);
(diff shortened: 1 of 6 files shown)


jonwetzel:
Critical bug found.

1. Click add equation
2. Select a schema
3. Click "Done"  At first the node does not close.  Incorrect equation syntax appears in the box.  (Not sure if this is expected, will talk with Kurt.)  
4. Click "Done" again
5. Click "Add Equation" and check console for error.
After this "Add Equation" no longer works and clicking on the nodes does not open the node editor due to console error:

```
Uncaught Error: parse error [column 1]: parity
    at Parser.error_parsing (parser.js?1536168231565:778)
    at Parser.parse (parser.js?1536168231565:765)
    at Function.Parser.parse (parser.js?1536168231565:555)
    at Object.<anonymous> (equation.js?1536168231565:571)
    at Object.forEach (array.js?1536168231565:247)
    at Object.getVariableStrings (equation.js?1536168231565:570)
    at Object.fillVariableNames (con-author.js?1536168231565:1408)
    at Object.initialControlSettings (con-author.js?1536168231565:928)
    at Object.showNodeEditor (controller.js?1536168231565:558)
    at Object.advice (aspect.js?1536168231565:114)```

**Comment by jonwetzel** [comment]
@riteshsamala Did you make any changes to the math solver for this?  I pulled the latest but it's still not working.

**Comment by jonwetzel** [comment]
@riteshsamala Found another critical bug.  I filled out an equation properly but it gives "Incorrect equation syntax".

eq-editor-bug2

**Comment by jonwetzel** [comment]
@riteshsamala As discussed, make the following fix: 
- [ ] Make it so the node editor done button doesn't close the node if the equation has a syntax error--instead display the message in the message box and leave the editor open so the author can fix it.

File: www/js/con-author.js [hunk 1]
```diff
@@ -588,6 +588,14 @@ define([
 			var logObj = {};
 			if(model && model == "authored"){
 				var directives = [];
+				//before parsing the equation we need to check if all variables are present and none of the variable fields is left empty
```
jonwetzel: Looks good, but I found a bug related to this function--if the user clicks "Delete Node" this check still triggers.  The equation done handler shouldn't need to be called on delete I think...

www/js/equation.js
```diff
@@ -556,6 +556,28 @@ define([
...
+			if(eqs.length != 2){
+				throw new Error("Wrong number of equal to symbols in the equation");
+			}
+			var expressions = [];
+			try{
+				array.forEach(eqs, function(eq, count){
+					expressions[count] = Parser.parse(eq);
+				}, this);
+			}catch(e){
+				throw e;
```
jonwetzel: Why a try-catch block if we're only just going to throw the error anyway?

jonwetzel: Logging the error sounds good to me.

riteshsamala: Can we push all the logging to a separate task later and leave a note there now that we have more prioritized tasks ?

jonwetzel: Sure.

jonwetzel:
New bug found:

* Add a DRT where D1 = R1 * T1
* Click Done
* Click on DRT eqn to edit it
* Change D1 to Q1
* Click Done

The equation is valid, but the node editor does not close on the first click.

jonwetzel:
New bug found:  Deleting an equation doesn't work

* Add a DRT where D1 = R1 * T1
* Click Done
* Click on DRT eqn to edit it
* Click Delete Node

Node editor closes but the node is not deleted.

jonwetzel:
We tried but neither of the new bugs were reproducible on other machines.  (This happened on the one in BYAC 110)

Issue #362, opened 2018-07-06
Schema Editing Mode - Step 1: Creation


jonwetzel:
New major mode.  Like Author mode, but with a few differences:

1. Add the mode.  Call it EDITOR.  Also include in login.html.
2. All fields related to teachers creating problems to student should be hidden.  This includes:
- [ ] sought quantity (also disable the check for this that comes when the user clicks "Done" in the main menu)
- [ ] Optionality
- [ ] Given to student

PR #396, opened 2018-09-18
schema editing mode v1

riteshsamala:
card #362


riteshsamala:
These changes work well, even I found the reference to edit auth lib was throwing that. I removed it and all the functionality looks well.

PR #402, opened 2018-09-20
schema reconstruction: equation editor UI

riteshsamala:


www/js/model.js
```diff
@@ -191,11 +191,12 @@ define([
 				// Summary: returns a list of all distinct entities author has entered for the model
 				var entityList = new Array();
 				array.forEach(this.authored.getNodes(), function(node){
-					if(node.entity && array.indexOf(entityList, node.entity) == -1){
-						entityList.push(node.entity);
+					//out of the list of entities, we pick the first one
```
jonwetzel: @riteshsamala I think the subtask was confusing on this one.  The entity dropdown must include ALL the entities the author entered in the entire model, with duplicates removed--not the first entity for each node.  So if the author has three nodes with schema and entities lists of <DRT: Bob;Bob's Plane;Plane> <Tailwind: Bob;Bob's Plane;Plane> <DRT: Jim;Jim's Car;Car> then the list of entities given to the student would be [Bob, Bob's Plane, Plane, Jim, Jim's Car, Car].

www/js/con-student.js
```diff
@@ -229,6 +235,40 @@ define([
 			this.applyDirectives(this._PM.processAnswer(this.currentID, 'units', unit));
 			// Logging
 		},
+		/* handle schema for student mode
+		*/
+		handleSchemas: function(schema){
+			var message;
+			var returnObj;
+			//case 1: if the student has equalled the number of schemas authored, throw an error in message box
+			if(!(this._model.authored.getTotalSchemaCount() >  this._model.student.getTotalSchemaCount())){
```
jonwetzel: This should be checking the number of the specific schema, not all schemas in the problem.  So you can use getGivenSchemaCount for this check to see if the student has too many.

www/js/con-student.js
```diff
@@ -266,7 +331,67 @@ define([
...
+				array.forEach(authorEntities, function(ent){
+					var obj = {value: ent, label: ent};
+					entityWid.addOption(obj);
+				}, this);
+
+				//disable the description and equation fields
+				descriptionWidget.set('disabled', true);
+				equationWidget.set('disabled', true);	
+			}
+			/*
```
jonwetzel: @riteshsamala  Is the plan to uncomment this after we get the new controls in?  The description part is obsolete but quantity node variable name, type, and other fields should all work the same as before...

jonwetzel:
@riteshsamala Is there any useful code left on this branch?

riteshsamala:
No @jonwetzel.

PR #409, opened 2018-10-02
Sharing sedit

riteshsamala:


jonwetzel:
Merging this, as we used this branch in class this term successfully.

Issue #361, opened 2018-07-06
Schema Reconstruction Mode


jonwetzel:
## Equation Node Editor Fields
- [x] Make it so the schema, entity, and variable name fields appear in the equation editor.  Hide the old equation input controls for now.  Use controller.js for any shared code with con-author.
- [x] The entity field should be a dropdown, populated with all the entities the author has entered from the entire mdoel (with duplicates removed).
- [x] The schema field should allow the student to try 2 times, and then the system should give them a schema from a random equation that is missing from the model in yellow.  (E.g. if the author model has 2 PWs and 1 DRT and the student model has one PW so far, randomly give either a PW or a DRT.  Count schemas in both models and compare to check.)
- [x] The schema field should be disabled once it is green or yellow.
- [x] The entity field should not be editable until the schema field is green or yellow.
- [x] The entity field should allow the student to try 2 times, and then the system should give them a random correct entity that would match an equation in the author's model that is missing from the student's model of that schema type.
- [x] The entity field should be disabled once it is green or yellow.

- [x] Equation description field should work as in author mode (move code to controller.js as needed)
- [ ] If the description field is colored, it should be green if both schema and entity are green.  If either of those is yellow, it should be yellow.
- [x] Variable names should not include the default new node name (e.g. W1)  If the student does not want to use an existing node, they must type in their own variable name.
- [x] Disable or hide the variable boxes until BOTH the schema and entity are correct.

## Quantity Node Editor Fields
When students opens a variable node, there are two cases:  A) We know what quantity in the author’s model the variable refers to (e.g., because the variable fills a slot of DRT) or B) we know only what set of quantities it could refer to (e.g., because it fills one of the Part slots of PW or average).  We will have TopoMath fill in the description for case A and have the student select a description for case B.

- [ ] Quantity description field should be a select box in student mode.  If the variable is not a part slot in a part-whole or average schema, its quantity can be copied from the author's model immediately when the node is created by the equation.  Otherwise, when the node opens the student will have to select from all the author's parts for that schema-entity equation (e.g. all the parts of the part whole, or all the parts of an average).
- [ ] Quantity type field -- make sure that in student mode this defaults to unset instead of unknown

## Equation Node Editor Feedback

Schema field feedback:
- [x] If the student has entered a schema that isn’t use in the author’s model, then they get feedback saying, “The author’s model doesn’t uses any [schema name] schemas.”
[...]

PR #437, opened 2019-03-19
Schema Reconstruction Mode (new version)

jonwetzel:
Issues we should fix before release:

- [x] **Issue A** [Icons are missing on quantity nodes](https://github.com/Dragoon-Lab/topomath/pull/437#issuecomment-474198927) 
- [x] **Issue B** [User can rename quantity to a name that is already in use](https://github.com/Dragoon-Lab/topomath/pull/437#issuecomment-474201235)
- [x] **Issue C** [User can reuse a variable in multiple slots, breaking equation editor](https://github.com/Dragoon-Lab/topomath/pull/437#issuecomment-474211166)
- [x] **Issue D** [Incorrect answers in node become yellow immediately when the equation node is re-opened.](https://github.com/Dragoon-Lab/topomath/pull/437#issuecomment-474692437)
- [x] **Issue E** Quantity editor should hide/show value field and always show units field.
- [x] **Issue F** Delete node fails on an equation with only the schema and entity set.
- [x] **Issue G** [Deleting an equation node with schema and entity set (green or yellow) means the next time that equation is created the user can no longer set the variables.](https://github.com/Dragoon-Lab/topomath/pull/437#issuecomment-475131114)
- [x] **Issue H** Entity help tooltip for reconstruction is missing.  Make it: "Select the entity this schema is about."
- [x] **Issue I** An equation node looks correct in canvas model (green check icon) when it is actually incorrect.
- [x] **Issue J** [Can't add variables to an equation after leaving the previous equation with missing variables.](https://github.com/Dragoon-Lab/topomath/pull/437#issuecomment-475147335)
- [x] **Issue K** A user should not be able to delete an equation where the schema and entity are set (green/yellow).

Issues we look at after release:

- When the quantity editor opens, let's change that dialog to not automatically put the cursor in the variable name box.  It changes the color from green to white which could make the user think they didn't get the variable name correct.  
- After deleting an equation, the equation counter does not decrement by 1 (it should).  Check behavior of quantity delete as well.
- On a touch screen, there is a console error when user clicks on the quantity type radio button directly.  (Workaround: click on the text instead).
- When the student gets yellow on the schema field, the message feedback is "You have entered a valid schema", but it should say "The correct answer has been given." like in other fields.
- If the student hits "Done" on the equation node with only a schema and entity, it gets the red x icon.  It should be incomplete green instead, since they haven't gotten anything wrong yet.
- For quantity nodes from schemas with ambiguous slots, when the description changes, the student model node should change which author node it matches with.

### DIFF
--- a/www/index.php
+++ b/www/index.php
@@ -307,6 +307,13 @@
 					<div id="qtyDescriptionQuestionMark" class="questionMark"></div>
 					<label for="qtyDescriptionInputbox">Description</label>
 				</span>
+				<input id="qtyDescriptionInputbox" data-dojo-type="dijit/form/TextBox">
+			</div> 
+			<div id="qtyDescriptionInputboxContainerStudent" class="fieldgroup" style="display: none;">
+				<span class="fixedwidth">
+					<div id="qtyDescriptionQuestionMark" class="questionMark"></div>
+					<label for="qtyDescriptionInputboxStudent">Description</label>
+				</span>
 				<select id="qtyDescriptionInputboxStudent" data-dojo-type="dijit/form/Select">
 					<option value='defaultSelect'>--Select--</option>
 					<option value='test1'>test1</option>
--- a/www/index.php
+++ b/www/index.php
@@ -235,6 +235,16 @@
 				<input id="entityInputbox" data-dojo-type="dijit/form/TextBox">
 			</div>
 
+			<div id="entitySelectorStudentContainer" class="fieldgroup" style="display: none;">
+				<span class="fixedwidth">
+					<div id="entityDescriptionQuestionMarkStudent" class="questionMark"></div>
+					<label for="entitySelectorStudent">Entity</label>
+				</span>
+				<select id="entitySelectorStudent" data-dojo-type="dijit/form/Select">
+					<option value='defaultSelect'>Select an entity</option>
+				</select>
+			</div>
+
 			<div class="fieldgroup">
 
 				<div id="descriptionInputboxContainer" class="fieldgroup" style="display: none;">
--- a/www/index.php
+++ b/www/index.php
@@ -6,7 +6,7 @@
 	<meta http-equiv="X-UA-Compatible" content="IE=EDGE" />
 	<title>TopoMath</title>
 	<link rel="shortcut icon" href="images/favicon.ico" type="image/x-icon"/>
-	<link rel="stylesheet" href="css/fontawesome.min.css">
+	<link rel="stylesheet" href="//maxcdn.bootstrapcdn.com/font-awesome/4.3.0/css/font-awesome.min.css">
 	<link rel="stylesheet" href="./dijit/themes/claro/claro.css">
 
 	<script type="text/javascript">
--- a/www/index.php
+++ b/www/index.php
@@ -307,7 +307,11 @@
 					<div id="qtyDescriptionQuestionMark" class="questionMark"></div>
 					<label for="qtyDescriptionInputbox">Description</label>
 				</span>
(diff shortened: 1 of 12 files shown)


jonwetzel:
PR Issue A:

After finishing this node like so:
image


The icon border turns yellow but the box is still empty, as if it is incomplete:
image

jonwetzel:
# PR Issue B
I am able to rename a variable name to an existing variable name and it turns green.  This should result in a red as [described in the task](https://github.com/dragoon-lab/topomath/issues/361).

image

Closing the node should result in the variable reverting back to its original name, but instead the duplicate name is saved.

jonwetzel:
# PR Issue C
The user can use the same variable in multiple slots, and then they can't open the node editor any more.

Suggested fix: prevent the user from entering the same variable in multiple slots.

jonwetzel:
# Issue D
After getting the value wrong once and closing the node, the next time the node is opened it turns yellow before the user does anything.

### Steps 
1. Open problem "DRT Bullet" from "tca-private" folder in "non-class-models" section in student mode, with restart problem on. 
1. Create an equation, DRT for bullet 
1. (This step I did but should be optional, I think) Finish DBullet and TBullet as instructed in the topomath curriculum, chapter 1, first example with screenshots.
1. Open quantity RBullet
1. Choose Parameter
1. Set units to m/s
1. Enter 860 as the Value
1. Click Done.  (Node has red x now as expected)
1. Open RBullet again.  Looks like the following screenshot:

image

The value is yellow, but it should still be red with value of 860.

jonwetzel:
# Issue G
Deleting an equation after selecting the schema and quantity stops the user from being able to complete it in the future.  (Reconstruction)

1. Start a problem in reconstruction mode (e.g. DRT Bullet)
1. Click Add equation
1. Select a correct schema and entity
1. Delete the node
1. Click "Add Equation" again
1. Enter some variable names for all fields
1. Click Done.  The node closes, but the equation is still empty and no quantity nodes are created. (bug)

Possible fix: Once the schema and entity are both set (green or yellow), don't allow the user to delete the equation any more.

jonwetzel:
# Issue I
Equation has incorrect variables, but it has the green checkmark icon:

Open problem The Matrix in chapter 1 problems in reconstruction mode.  Make Neo's DRT first.  Then create the bullet DRT and use the same variables as your Neo DRT in all three slots.  No need to complete the quantity nodes first.  Result:

image

Open the node.  Actually two variables are incorrect:
image

So the node should have a red x instead of the green checkmark.

The bug continues: Opening the node again will cause the fields to appear yellow (that's Issue D), but unlike in Issue D they won't get auto corrected.  We need to ensure that the icon is consistent with the state of the node after it is closed.

jonwetzel:
# Issue J
Adding variables to an equation after leaving another equation with no variables doesn't work (console error).

image

Steps:
1. Open a problem with multiple equations.  I used "Bike Race" from problem set 1.
1. Add an equation and set the schema and entity.  Click "Done" without entering any variables.
1. Add another equation and set the schema and entity, then fill out the variables without closing the node editor.  
1. Click "Done" and check console for error.  Only the first variable appears on the canvas.

This only happens if the equation made before this one was made without its variables.  If you make the first equation with its variables, the second doesn't error during its autocreate step.

One idea for fix: Don't let the user close the node editor if a variable slot is blank, just like we don't let them close it if there are duplicates.

jonwetzel:
# Issue L
Variable slots become enabled even if the entity selected is incorrect.  That could lead to weird situations.  They should remain hidden until the entity is correct or demo.

Example:
image

jonwetzel:
Last commit fixes a bug where the quantity description field was green even before the user selects one, and it wouldn't be green the next time they opened the node.

PR #445, opened 2019-03-25
Reconstruction Bug Fixes

jonwetzel:
Testing bug fixes for ambiguous schema slots (e.g. parts of part whole)

A good test problem is "Age Puzzle" in non-class-models, folder "chapter 3 quiz extras-tca".

### DIFF
--- a/www/js/con-student.js
+++ b/www/js/con-student.js
@@ -53,6 +53,7 @@ define([
 			this.studentControls();
 			this.variableUpdateBySystem = false;
 			this.deleteNodeActivated = false;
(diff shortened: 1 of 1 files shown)


jonwetzel:
# Issue A
The units field is enabled before the description is selected.  The type field is disabled, but it's hard to tell that it is.  

Suggested fix: Hide both the units and type until the description is correct or demo'd.

jonwetzel:
# Issue B
Multiplication schemas (M2, M3, M4) should be in the ambiguous category too.

Issue #444, opened 2019-03-22
Jack and Jill DRT-- user can create multiple Jack DRTs?


jonwetzel:
Need to figure out how this happens.

Issue #443, opened 2019-03-22
Extra Incorrect equations become undeletable 


jonwetzel:
In reconstruction, add all the equations required.  Then add an extra equation, and select a schema and hit done.  If you close the node editor, then the node can no longer be deleted.  It shows up as green when you open the node.

PR #448, opened 2019-03-25
topomath #443, #444

riteshsamala:


jonwetzel:
#Issue A
This is a low priority issue for the backlog.  It may or may not be from this PR.

1. Add an equation and select any schema that is not in the problem.  
1. Select "Choose a Schema"
1. Again, select any schema that is not in the problem.   It changes to a valid schema, but it is red instead of yellow.

image

jonwetzel:
# Issue B

The first time we close an incorrect equation node, it cannot be deleted with right-click.

1. Add a new equation node.
1. Select an incorrect schema (get red)
1. Click Done to close the equation node
1. Right click on the equation node
Delete should be enabled, but it is not.

jonwetzel:
# Issue C

1. Click add equation
1. Click "Done"
1. Open the equation again
1. "Select a schema" is green and cannot be clicked

Minor issue.  Workaround: delete the node.

jonwetzel:
@riteshsamala Looks good, I'll merge this and put the issues on the backlog.

Issue #451, opened 2019-03-26
The first time we close an incorrect equation node, it cannot be deleted with right-click.


jonwetzel:
1. Add a new equation node.
1. Select an incorrect schema (get red)
1. Click Done to close the equation node
1. Right click on the equation node
Delete should be enabled, but it is not.

Issue #446, opened 2019-03-25
A red variable slot should be white after the user edits it


jonwetzel:
1. Get a slot variable wrong on an equation node
2. Re-open the equation and edit the red slot to be correct
3. Click another spot in the node editor to remove focus from the slot

The slot should remain white, but it becomes red again.  This leads students to think they got the answer wrong, when it actually hasn't been checked yet.  Make it stay white after editing until the user closes the node editor again.

Issue #441, opened 2019-03-22
Sometimes author nodes are getting autocreated into the student model, invisibly


jonwetzel:
Need to figure out how to reproduce it.

PR #453, opened 2019-03-26
Bugfix #451 #446 and #455

riteshsamala:
Status:

- [x] #446 Fixed
- [x] #451 Fixed
- [x] Issue A Fixed
- [x] #455 Fixed

### DIFF
--- a/www/js/con-student.js
+++ b/www/js/con-student.js
@@ -757,15 +757,22 @@ define([
 				var original_id = this._model.student.getAuthoredID(this.currentID);
 				var original_eq = this._model.authored.getEquation(original_id);
 				var original_Ar = expression.getVariableStrings(original_eq);
+				var original_RightAr = expression.getRightSideEquationStrings(original_eq);
 				var current_Ar = expression.getVariableStrings(this._model.student.getEquation(this.currentID));
+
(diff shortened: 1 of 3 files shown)


jonwetzel:
# Issue A

1. Add an equation node and fill it out complete correctly.
1. Click done (Should get gold star)
1. Right click on the equation.  
You should not be able to delete it, but you can.

jonwetzel:
# Issue B
This issue is probably not unique to this PR--it should probably be merged into #441.  (Will confirm with more testing.)

1. Open Age Puzzle problem
1. Complete P3W eqn
1. Add M2 eqn Barbara + Emma and for slots enter W, MBarbara, MJohn (all 3 incorrect)
1. Click Done.  Red x comes as expected.
1. Open that M2 eqn again.  All slots are red as expected.
1. Change the slots to  MBarabara (correct), MJohn (incorrect), M1 (correct)  Click "Done".  Two bugs follow:

First, the canvas node is yellow, but the equation is not updated to the new equation.
image

Second, even though "M1" was correct, Multiplier1 is added instead.  This node does not show up on the canvas.  I believe this is the cause of bug #441.  
image

jonwetzel:
# Update to #455 

In an M2 eqn, incorrect feedback is given if new variable is in the wrong slot, even though order should not matter for this: 
image
vs
image

Issue #477, opened 2019-06-07
Edit Schema table


vanlehn:
See changes in the attached table.  Rows in the schema table that are not mentioned in this table are to be left alone 
 [https://app.zenhub.com/files/72044530/7d32b6fe-e4ae-4a7f-b53e-ee6ad55b05d0/download](https://app.zenhub.com/files/72044530/7d32b6fe-e4ae-4a7f-b53e-ee6ad55b05d0/download)

Issue #476, opened 2019-06-07
AutoGenerate variable names to fill schema slots


vanlehn:
Concatenate the variable name in the schema table with the first word in the relationship field.  Use this text as if the user had typed it in.

Issue #474, opened 2019-06-07
Alien quantity nodes should not have description field set


vanlehn:
Test on Desalinization in chapter 3 quiz extra-tca

jonwetzel:
@riteshsamala Can you tell me the steps to test this?  I'm not sure what nodes are being referred to...

riteshsamala:
@jonwetzel there was a case where in alien nodes where the description was being set because of the bug, In desalization problem, when one of the mixture nodes had a wrong slot entry, eventually after the system corrected, these became alien nodes but still had a description before the bug (the description field was turning green).

Issue #472, opened 2019-06-07
Suppress units when switch is set


vanlehn:
In order to save user interface actions, when a new switch is set to true, the units field does not appear in any quantity node.  The check for completeness, correctness, etc. must be modified as well.

PR #483, opened 2019-06-18
Schema recon2

riteshsamala:
issues addressed
https://github.com/Dragoon-Lab/topomath/issues/454
https://github.com/Dragoon-Lab/topomath/issues/471
https://github.com/Dragoon-Lab/topomath/issues/474
https://github.com/Dragoon-Lab/topomath/issues/475
https://github.com/Dragoon-Lab/topomath/issues/473
https://github.com/Dragoon-Lab/topomath/issues/472
https://github.com/Dragoon-Lab/topomath/issues/476
https://github.com/Dragoon-Lab/topomath/issues/477
https://github.com/Dragoon-Lab/topomath/issues/480
https://github.com/Dragoon-Lab/topomath/issues/479
last commit : couple of bugs related to 480 ( no issues for added for them, but need these to test 480 and also 479).

### DIFF
--- a/www/index.php
+++ b/www/index.php
@@ -339,7 +339,7 @@
 				</div>
 
 				<div id = "unitsSelectorContainerStudent" class="fieldgroup" style="display: none">
-					<label for="unitsSelectorStudent">Units</label>
+					<label for="unitsSelectorStudent" id="unitsSelectorStudentLabel">Units</label>
 					<select id="unitsSelectorStudent" data-dojo-type="dijit/form/Select">
 						<option value='defaultSelect'>No Units</option>
 					</select>
--- a/www/js/con-student.js
+++ b/www/js/con-student.js
@@ -955,6 +955,11 @@ define([
 			var authoredID = this.setNodeDescription(node.id, node.variable);
 			if(authoredID){
 				this.updateInputNode(node.id, node.variable);
+				//if the give parameters switch is on, according to the latest design update refer card #473, all the unknown nodes should be automatically set along with the status
+				if(sessionStorage.getItem("draggp") == "on"){
+					this._model.student.setVariableType(node.id, "unknown");
+					this._model.student.setStatus(node.id, "variableType" , {"disabled":true,"status":"correct"});
+				}
 				var canHaveDeleteNode = this.canHaveDeleteNode(node.id);
 				this._model.student.setCanDelete(node.id,canHaveDeleteNode);
 				this.updateNodeView(this._model.active.getNode(node.id));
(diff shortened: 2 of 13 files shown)


jonwetzel:
@riteshsamala Author mode is erroring out without loading.  Any ideas?

image

riteshsamala:
Yeah I fixed this bug, the code is on api branch ... I also made a commit to this branch as we are testing it currently. Will omit the commit on api branch when I raise a PR. Please test now

jonwetzel:
Testing progress:

- [ ] #454
- [ ] #471
- [ ] #474
- [ ] #475
- [ ] #473
- [ ] #472
- [ ] #476
- [ ] #477
- [ ] #480
- [ ] #479

riteshsamala:
Interesting, checking it now !

jonwetzel:
Testing progress:

- [x] #454 Working in author and student.  6/27
- [ ] #471 and #474 Found a bug (see review above)
- [x] #475 
- [x] #473 Fixed 
- [ ] #472 See comment https://github.com/Dragoon-Lab/topomath/commit/a8f9161940e6ccda13be0cfd7bd936b607e838cb#commitcomment-34091727
- [x] #476  Working 6/27
- [x] #477 and #482 Schema table is updated properly.
- [ ] #480
- [ ] #479 See suggestion https://github.com/Dragoon-Lab/topomath/commit/628a45a4129610ba67ff1fe97a939c6fbeecb865

Test problem: test1/login.html/test620

riteshsamala:
> If found a bug: when skip units is turned off, units says it's incorrect even when the user selects the right answer.
> image
> 
> Double check to make sure units works when not skipped.

In the author model, pdrinking does not have any units set as far as this case in concerned, will test the following bugs and see if this can be reproduced.

jonwetzel:
@riteshsamala Giving DRTs still doesn't work for [the same problem I was testing earlier](https://github.com/Dragoon-Lab/topomath/pull/483#pullrequestreview-255436157).  For some reason it creates an alien node instead of sharing the variable between the two DRTs.

Starting that problem with restart problem and give DRTs checked gives:
image
The lower DRT should share the TJack variable, but it tries to create a TJill instead.

jonwetzel:
Confirmed the bug with #472 (skip units) is fixed.  Units work in all 4 cross-cases of author-added-or-not x skip-units-or-not.

jonwetzel:
Confirmed #471 and #474 bug is fixed.

jonwetzel:
@vanlehn Aside from the bug 3 comments up on given-DRTs-with-a-shared-variable and the reappearance of #465, it looks like everything else is fixed.  New version is up on devel for your testing if you like.

riteshsamala:
> For

Hi jon, 
From what I understood with discussions from kurt, TJill is the ideal auto generated name and user has to correct it later(even when drts are given). @vanlehn prof, can you please confirm this.

vanlehn:
Yes, that’s correct.
--Kurt

PR #491, opened 2019-06-27
api changes to accommodate new LMS

riteshsamala:


riteshsamala:
@jonwetzel, turns out the fix was needed on LMS side. I updated LMS code, please test now.

PR #498, opened 2019-07-18
Issue 458

sachingrover211:
Issue completed --
[1] Error scenario setup for static solution singular matrix.
[2] Error handled on rendering side by changing NaN values to the extrapolated solution from the rest of the values.
[3] Rendering comes correct for the NaN values, tested by hand calculation for a small dummy problem as well as the problem mentioned in issue 436.

### DIFF
--- a/www/js/chart.js
+++ b/www/js/chart.js
@@ -169,6 +169,14 @@ define([
 			return series;
 		},
 
+        formatDashedSeries: function(stroke){
+            array.forEach(stroke, function(s){
+                s["style"] = "ShortDash";
+            });
+
+            return stroke;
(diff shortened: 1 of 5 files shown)


www/js/chart.js
```diff
@@ -169,6 +169,14 @@ define([
 			return series;
 		},
 
+        formatDashedSeries: function(stroke){
+            array.forEach(stroke, function(s){
+                s["style"] = "ShortDash";
```
jonwetzel: I guess this was the only part that didn't work?  @sachingrover211 said he will make a fix later for the styling, which is fine.

Issue #495, opened 2019-07-16
Error msg when Done button pressed and alien nodes still present


vanlehn:
When the Done button is pressed in immediate feedback mode and alien nodes are still present, then the error message currently complains about incomplete nodes.  Change it to say, "Some alien nodes need to be deleted."

jonwetzel:
Merged in PR #499

Issue #478, opened 2019-06-10
Remove Graph button.  Relabel Table button to be Solve.


vanlehn:
Users can still get to the Graph using the tabs i n the Table window, but using the graph window will no longer be taught.  The graph window will just be ignored in the current curriculum, but might be needed in future curricula.

jonwetzel:
https://github.com/Dragoon-Lab/topomath/pull/499

Issue #465, opened 2019-05-07
Right click delete doesn't work on unused quantity nodes


jonwetzel:
1. Add an equation and enter a new variable name when an existing one should be used.  (Easy to reproduce if you do "Give parameters" so you don't have to make another equation to generate the variable first.)
1. Click "Done".  The extra node you entered is autocreated.
1. Open the equation again
1. Correct the error and click done.
1. The extra node is still there, time to delete it.  Right click it and... delete is disabled.
Delete should be enabled.

Workaround: Open the node.  It can be deleted from within the node.  

Isn't the same logic used to disable/enable delete inside the node editor and in the right click menu?  Why doesn't it work?

jonwetzel:
@riteshsamala Which branch is this fix on?  I was able to reproduce this in the following scenario:

1. Create an equation in author mode
2. Create a second equation without reusing any of the nodes from equation 1.
3. Edit equation 2, and use the dropdown to change one of the nodes to be a different one.
4

jonwetzel:
Also, quantity nodes which belong to equations are able to be deleted from inside the node editor.  Instead the button should be disabled and the tooltip explains why they can't delete if they mouseover it.

jonwetzel:
@riteshsamala Still broken in schema_Recon2 after merging latest master.

jonwetzel:
This bug is back.

riteshsamala:
Ok, testing now

Issue #464, opened 2019-05-07
When the student tries to add an equation but no more equations are needed, TopoMath should tell them they don't need more equations


jonwetzel:
On clicking Add Equation, if the student model has equation nodes for all the schemas in the author's model already (e.g. right number of every schema, entities and variables don't need to be correct yet) then instead of opening the equation editor we just show a modal message that says "You don't need any more equations." with an "Ok" button to dismiss it.

jonwetzel:
One idea: On clicking Add Equation, if the student model has all the equations already, instead of opening the equation editor we just show a modal message that says "You don't need any more equations."

jonwetzel:
Converting this design task to user story.

jonwetzel:
> On clicking Add Equation, if the student model has equation nodes for all the schemas in the author's model already (e.g. right number of every schema, entities and variables don't need to be correct yet) then instead of opening the equation editor we just show a modal message that says "You don't need any more equations." with an "Ok" button to dismiss it.

This design fails if the student adds a extra equations before setting the schemas.  Just count the number of equations in the Author's model, and if the student has that many or more show the modal.

jonwetzel:
https://github.com/Dragoon-Lab/topomath/pull/499

PR #499, opened 2019-07-18
Issues #478 , #464, #465, and #495

riteshsamala:
Addresses issues #478 , #464, #465, and #495.

### DIFF
--- a/www/js/draw-model.js
+++ b/www/js/draw-model.js
@@ -568,6 +568,9 @@ define([
 				if(connection.sourceId == nodeID || connection.targetId == nodeID){
 					try{
 						this._instance.detach(connection);
+						var lostNode = connection.sourceId === nodeID ? connection.targetId : connection.sourceId;
+						//if the updated connection points to an alien node (no authoredID), then update the candelete field and delNodeTag which facilitates right click delete
+						this.updateDeleteLabel(lostNode);
 					}
 					catch (err){
 						console.log("Error while detaching: " + err);
@@ -592,6 +595,14 @@ define([
 				array.forEach(arguments[1], lang.hitch(this, function(node){
 					if(node.name && node.description) this.addNodeDescription(node.id);
 				}));
+			}
+		},
+
+		updateDeleteLabel: function(lostNode){
+			//if the updated connection points to an alien node (no authoredID), then update the candelete field and delNodeTag which facilitates right click delete
+			if(!this._model.getAuthoredID(lostNode)){
+				this._model.setCanDelete(lostNode, true);
+				dijit.byId(lostNode+"delNodeTag").set("disabled", false);
 			}
 		}
 	});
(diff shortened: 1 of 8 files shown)


riteshsamala:
Also addresses #495

www/js/model.js
```diff
@@ -243,6 +243,25 @@ define([
 						}, this);
 				return flag ? true : false;
 			},
+			equationNodesMatchesGivenSolution: function(){
```
jonwetzel: This condition doesn't match the spec in #464.  It is checking if every author equation has been added to the student model and if all of them are complete.  It was just supposed to check if the student model has the right number of every schema.  So that created two bugs:

1) The user must fill in the entities and variables before the check will work.  (Fixed by only counting schemas, ignore entities and variables)
2) If the student has one or more incomplete equations , the check fails even if all the equations are present.  (Fixed by removing isComplete check.)  

However the second bug is also in the 464 design.  If the user clicks "Add Equation" a bunch of times without setting the schema, they can add more than the max number of equations and end up in the same situation that 464 was trying to fix.  So instead of using the old design I wrote in #464, make the following changes in this pull request:

1. Rename the function to `checkEquationNodesDone` to `hasEnoughEquationNodes`
1. Change that function to just count the number of equations in the Author's model and count the number of equations in the student model.  If the student's count is >= the Author's count, show the message and don't add the equation.  If the counting function exists already, do reuse it.

www/js/main.js
```diff
@@ -314,6 +314,18 @@ define([
 			//next step is to add action to add equation
 			menu.add("createEquationNodeButton", function(e){
 				event.stop(e);
+				//before a node is created, check if the model has enough equation nodes
+				//applies only to student mode
+				if(_model.isStudentMode()){
+					var allEquationNodesDone = controllerObject.checkEquationNodesDone();
+					if(allEquationNodesDone){
+						var buttons = [];
+						var title = 'Equation Nodes Complete!';
```
jonwetzel: Change this title to "Add Equation" (refers to the button that was clicked to cause the message)
```suggestion
						var title = 'Add Equation';
```

www/js/draw-model.js
```diff
@@ -568,6 +568,9 @@ define([
 				if(connection.sourceId == nodeID || connection.targetId == nodeID){
 					try{
 						this._instance.detach(connection);
+						var lostNode = connection.sourceId === nodeID ? connection.targetId : connection.sourceId;
+						//if the updated connection points to an alien node (no authoredID), then update the candelete field and delNodeTag which facilitates right click delete
+						this.updateDeleteLabel(lostNode);
```
jonwetzel: (No change required here, just a comment.)

It seems a little odd the draw-model would have to do this update instead of it being done in the model code (as the model should contain all the info required to do this update by itself).  But I guess that's because our model and view are too tightly coupled already.  Changing that would be a much, much bigger refactoring that we don't have time for right now, but let's keep it in mind for the future.

jonwetzel:
The fix for #465 doesn't work for some reason, even though it did the firs time I tested this PR.  Odd.  Merging so we will get the higher priority features/fixes, which do work!

Issue #513, opened 2019-07-26
When both parameters and DRTs are given and parameters have default names the DRT should use those parameters


jonwetzel:
In this example, the DRTs should re-use the existing nodes except when they are missing, like DNormal and DConstruction.  Instead the DRTs are getting new copies of every parameter like so:

![image.png](https://images.zenhubusercontent.com/5873c675b9dc3c62202498ef/83a3f5cf-70cc-4358-8b49-9b4880ec982e)

The student then can't change the equation to the correct node since it has the same name.

This happens in: 

- Myras Commute in v2 chapter 1 problems
- The Matrix in chapter 1 problems

jonwetzel:
This is a critical bug because it student is unable to finish the problem without deleting the given DRTs or renaming the nodes first.  Neither is intended.

riteshsamala:
working on it currently

jonwetzel:
After making the fix, test the following conditions:

With both DRTs and Params given:

1. A model that has parameters with only default names (e.g. "DNormal")  In this case there should be no alien nodes because the parameters all have default names.
1. A model that has no parameters with default names (e.g. all named things like "NormalDistance")  In this case, there should be alien nodes for all parameters because the parameters had non-default names.
1. A model that has some parameters with default names and some parameters with non-default names.  In this case, the DRTs should not make alien nodes for the parameters with default names and should make alien nodes for the parameters which had non-default names.

PR #515, opened 2019-07-29
When both parameters and DRTs are given and parameters have default n…

riteshsamala:
…ames the DRT should use those parameters #513


riteshsamala:
I could not reproduce this on my mac. Attached screen recording on slack.

PR #540, opened 2019-09-30
Dashboard issue

sachingrover211:
Null checks added on line 538 and 546. Cause due to wrong order of log messages sent/received due to internal race conditions or network issues.


jonwetzel:
Will merge after testing.

Issue #531, opened 2019-08-29
Swapping two variables to correct a mistake gets a yellow instead of green


jonwetzel:
Steps:

1. Open a problem.  We'll use "Life Points" in "chapter 2 problems" on tca in reconstruction mode, with Give parameters and Skip units checked (also restart problem).
1. Click Add equation (we'll get it wrong first)
1. Select "Compare" schema
1. Select "Points" entity
1. Set Bigger to Bob, Smaller to SPoints, and Difference to Diff
1. Click done.  Answer is incorrect as expected
1. Open equation again.  Bob and SPoints are red as expected.
1. Set Bigger to "SPoints" and Smaller to "Bob".  Click Done

At this point the selected answers are correct, so the equation should just be green.  Instead, it registers as incorrect and gives a yellow, autocreating a new answer for the "Bigger" field.

Issue #367, opened 2018-07-24
Author mode: Add feedback to variable slots


jonwetzel:
Since only alphanumeric characters can be entered, the feedback is simple: the combo box should turn red when the field is blank.  Blue otherwise.

PR #541, opened 2019-10-03
Schema recon2

riteshsamala:
Handles 
issues topomath #367 , #531

### DIFF
--- a/www/css/global.css
+++ b/www/css/global.css
@@ -333,6 +333,21 @@ input[type=radio]{
   box-shadow: 2px 2px 2px darkolivegreen;
   border-color: green;
 }
+.topomath-feedback.fa-gift.perfect{
+  color: rgb(153, 134, 9);
+  box-shadow: 2px 2px 2px darkolivegreen;
+  border-color: green;
+}
+.topomath-feedback.fa-gift.incorrect{
+  color: rgb(153, 134, 9);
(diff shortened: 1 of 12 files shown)


jonwetzel:
@riteshsamala 

Found a bug with this branch on tdevel.  If you get an equation wrong twice, the node editor doesn't close on your second attempt.  It's erroring out in the console with "Assertrion failed: dom.byId failed for <node name>" and then "Uncaught TypeError: Cannot read property 'constructor' of null".

Steps:
1) Open "Peregrine Falcon" with parameters given and units skipped.  section non-class-models, folder "chapter 1 problems-tca".
2) Create a new equation and select DRT and Dive.
3) Don't change the node names, just hit Done.  The node closes and the incorrect nodes are created.  The equation is red as expected.
4) Open the equation again.  Change DDive to Distance (the correct answer).
5) Click Done.  Bug happens.

If this doesn't happen in the branch where students get three attempts, it check it after merging with this one.

jonwetzel:
@riteshsamala That fixed the bug, but now https://github.com/Dragoon-Lab/topomath/issues/543 doesn't work anymore (which was an earlier commit on this PR).

Follow the same steps as above, then check the model after closing the node.  The alien node is still present.

jonwetzel:
Tested thoroughly in trials Feb 15 and before.

